PR #113, opened 2026-04-08
Move near signing to rust

0xh3rman:


ios/Packages/Signer/Sources/Signer.swift
```diff
@@ -81,7 +81,7 @@ public struct Signer: Sendable {
         case .tron: TronSigner()
         case .bitcoin: BitcoinSigner()
         case .xrp: XrpSigner()
-        case .near: NearSigner()
+        case .near: ChainSigner(chain: chain)
```
gemcoder21: same here, move to all cases for `ChainSigner`

gemcoder21 (approved):

PR #120, opened 2026-04-08
Add Xcode pre-build script for Gemstone XCFramework

0xh3rman:
- Add build-gemstone.sh that hashes Rust sources and only rebuilds when changed, similar to Android Gradle inputs/outputs
- Add Run Script Build Phase to Gem target before Sources
- Disable ENABLE_USER_SCRIPT_SANDBOXING (script needs access to ../core and cargo)

image image

Need to run jun once for SPM resolution


ios/scripts/build-gemstone.sh
```diff
@@ -0,0 +1,43 @@
...
+if [ "${CONFIGURATION:-Debug}" = "Release" ]; then
+    BUILD_MODE="release"
+else
+    BUILD_MODE=""
+fi
+
+CACHE_DIR="$IOS_DIR/build/.gemstone-cache"
+HASH_FILE="$CACHE_DIR/sources-${BUILD_MODE:-debug}.hash"
+
+current_hash=$("$SCRIPT_DIR/gemstone-hash.sh" "$CORE_DIR")
```
gemcoder21: we have `generate-stone` method already - is it different? or why old one not removed?

gemcoder21 (approved):

gemcoder21:
Does this work well on first run? when you clone the repo too?

0xh3rman:
> Does this work well on first run? when you clone the repo too?

Need to manually run once, because SPM resolving happens before xcode build phases, this script doesn't have chance to run before that

PR #142, opened 2026-04-14
Navigate to existing wallet after import

DRadmir:
Set the imported wallet as current when the wallet already exists, so tapping Continue navigates the user to that wallet instead of staying on the previous one.

iOS: extract activateWallet() called for both new and existing cases.
Android: call sessionRepository.setWallet() for the existing case.


android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/wallets/PhraseAddressImportWalletService.kt
```diff
@@ -54,6 +54,7 @@ class PhraseAddressImportWalletService(
                 WalletType.PrivateKey -> handlePrivateKey(importType.chain!!, walletName, data)
             }
         } catch (err: ImportError.DuplicatedWallet) {
+            sessionRepository.setWallet(err.wallet)
```
gemcoder21: what does this do? strange to set Wallet on error?

gemdev111 (approved):

PR #149, opened 2026-04-15
Use Chain for DustThreshold and add UI info

### LINKED ISSUE #146: Add UTXO handling UX
We should handle in similar manner like we do for missing network fee, info button with explanation and link to 

Image

DRadmir:
Change ConfirmError.DustThreshold to carry a Chain instead of a string title so callers can access chain/asset data. Update ConfirmViewModel to construct DustThreshold with owner.chain. Adapt ConfirmScreen to use a short dust-threshold string resource. Add a DustThresholdInfo InfoSheetEntity (with icon, description and docs link) and map the error to that sheet in ConfirmErrorInfo. Also add tools:replace="android:dataExtractionRules" to the AndroidManifest entry.

Fix: https://github.com/gemwalletcom/wallet/issues/146

Screenshot_20260415_163554
Screenshot_20260415_163615

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -22,6 +22,7 @@
         android:name="com.gemwallet.android.App"
         android:exported="true"
         tools:node="merge"
+        tools:replace="android:dataExtractionRules"
         >
         <activity
             android:name="com.gemwallet.android.MainActivity"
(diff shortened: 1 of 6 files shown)


android/app/src/main/AndroidManifest.xml
gemcoder21, line 25: is this needed?

gemcoder21 (approved):
Needs handling on iOS too, it was not working in the screenshot

gemdev111 (approved):

PR #150, opened 2026-04-15
Move TON tx signing to rust

0xh3rman:
1. Move TON signing to Rust
2. Fix WalletConnect for TON
- - Configure session properties for TON
- - Read public key from keystore
- - Add SignAndSend output action type


gemcoder21 (approved):
🔥

ios/.swiftformat
```diff
@@ -2,4 +2,4 @@
 --exclude ./Packages/Primitives/Sources/Generated
 --exclude ./Packages/Gemstone/Package.swift
 --exclude ./Packages/Localization/Sources/Localized.swift
---exclude ./Packages/Localization/WidgetSources/Localized.swift
+--exclude ./Packages/Localization/WidgetSources/Localized.swift--test-case-name-format preserve
```
gemcoder21, line 5:
```suggestion
--exclude ./Packages/Localization/WidgetSources/Localized.swift
--test-case-name-format preserve
```

PR #211, opened 2026-04-23
Fix Android Monad stake delegation routing

DRadmir:
Scope delegation lookup by state so multiple positions on the same validator (Active / Activating / Deactivating / AwaitingWithdrawal) open their own detail screen instead of collapsing to the first match.

Also hide the orphaned "Manage" header when the actions list is empty, and keep watch wallets off the Confirm screen when tapping an AwaitingWithdrawal row.

### DIFF
--- a/android/app/src/main/kotlin/com/gemwallet/android/ui/navigation/routes/Stake.kt
+++ b/android/app/src/main/kotlin/com/gemwallet/android/ui/navigation/routes/Stake.kt
@@ -11,26 +11,27 @@ import com.gemwallet.android.ui.models.actions.AmountTransactionAction
 import com.gemwallet.android.features.earn.delegation.presents.DelegationScene
 import com.gemwallet.android.features.stake.presents.StakeScreen
 import com.wallet.core.primitives.AssetId
+import com.wallet.core.primitives.DelegationState
 import kotlinx.serialization.Serializable
 
 @Serializable
 data class StakeRoute(val assetId: String)
 
 @Serializable
-data class DelegationRoute(val validatorId: String, val delegationId: String)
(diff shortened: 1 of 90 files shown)


android/app/src/main/kotlin/com/gemwallet/android/ui/navigation/routes/Stake.kt
```diff
@@ -11,26 +11,27 @@ import com.gemwallet.android.ui.models.actions.AmountTransactionAction
...
 import com.wallet.core.primitives.AssetId
+import com.wallet.core.primitives.DelegationState
 import kotlinx.serialization.Serializable
 
 @Serializable
 data class StakeRoute(val assetId: String)
 
 @Serializable
-data class DelegationRoute(val validatorId: String, val delegationId: String)
+data class DelegationRoute(val validatorId: String, val delegationId: String, val state: String)
```
gemcoder21: why do we need to pass state to the route?

android/features/earn/delegation/viewmodels/src/main/kotlin/com/gemwallet/android/features/earn/delegation/viewmodels/DelegationViewModel.kt
```diff
@@ -221,7 +224,8 @@ class DelegationViewModel @Inject constructor(
             assetId = assetInfo.value?.asset?.id!!,
             txType = type,
             validatorId = delegation.value?.validator?.id,
-            delegationId = delegation.value?.base?.delegationId!!
+            delegationId = delegation.value?.base?.delegationId!!,
```
gemcoder21: can we avoid `!!` ?

gemcoder21 (approved):

PR #220, opened 2026-04-26
Fix: Android filter modal icon size

### LINKED ISSUE #180: Fix icon size on filtering
Image

jonathanchw:
Resize Networks and Types icons in the Activities filter modal to match iOS visual weight (`32dp` with rounded clip).

## Before
image

## After
image

Close #180

### DIFF
--- a/android/ui/src/main/kotlin/com/gemwallet/android/ui/components/filters/ActivitiesFilter.kt
+++ b/android/ui/src/main/kotlin/com/gemwallet/android/ui/components/filters/ActivitiesFilter.kt
@@ -5,6 +5,7 @@ import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.input.rememberTextFieldState
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.filled.Article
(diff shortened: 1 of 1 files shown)


jonathanchw:
ready for review. @gemcoder21

android/ui/src/main/kotlin/com/gemwallet/android/ui/components/filters/ActivitiesFilter.kt
```diff
@@ -60,7 +62,7 @@ fun TransactionsFilter(
                             text = R.string.settings_networks_title,
                             trailing = {
                                 Image(
-                                    modifier = Modifier.size(listItemIconSize),
+                                    modifier = Modifier.size(32.dp).clip(RoundedCornerShape(8.dp)),
```
gemcoder21, line 65: it should be using `iconSize` constant, can you check to make sure it's consistent with other settings options that it uses same PropertyItem? or list item? and avoid hardcoding values

android/ui/src/main/kotlin/com/gemwallet/android/ui/components/filters/ActivitiesFilter.kt
```diff
@@ -88,9 +90,9 @@ fun TransactionsFilter(
                             text = R.string.filter_types,
                             trailing = {
                                 Icon(
-                                    modifier = Modifier.size(listItemIconSize),
+                                    modifier = Modifier.size(32.dp),
                                     imageVector = Icons.AutoMirrored.Default.Article,
-                                    contentDescription = "networks filter"
+                                    contentDescription = "types filter"
```
gemcoder21, line 95:
```suggestion
                                    contentDescription = null
```

gemcoder21:
Fixed in c407ffac6df300948f0d4a99aa2a0d238efb5206

PR #246, opened 2026-04-30
Preload perpetual markets when enabling Perpetuals toggle

DRadmir:
Hyperliquid clearinghouseState positions hit FOREIGN KEY constraint when perpetuals/assets metadata is missing in DB


ios/Packages/FeatureServices/AppService/AppLifecycleService.swift
```diff
@@ -112,6 +118,17 @@ extension AppLifecycleService {
         }
     }
 
+    private func updatePerpetualMarkets() async {
```
gemdev111, line 121: Maybe move  out `updatePerpetualMarkets` from AppLifecycleService  somewhere to higher level?

PR #262, opened 2026-05-05
Android: Match checkmark size on Accept Terms screen

### LINKED ISSUE #257: Consolidate checkmark size on accept terms
Image
Image

DRadmir:
## Summary
- Resolves the size mismatch between checked and unchecked icons on the Accept Terms screen.
- `SelectionCheckmark` fills 100% of its box; Material's `Icons.Outlined.Circle` leaves ~17% padding inside its viewBox, so at the same `compactIconSize` the two icons rendered at visibly different diameters.
- Bumps the outer `Box` and the outlined icon to `smallIconSize` (24.dp) so the outlined circle's visible diameter matches the 20.dp filled checkmark.

Closes #257
Screenshot_1777969434


gemdev111 (approved):

gemcoder21:
Fixed on main already

PR #325, opened 2026-05-15
Split formatters by domain (iOS)

DRadmir:
Reduce CurrencyFormatter to pure fiat formatting. Extract domain-specific formatters with non-overlapping responsibilities:

- PercentFormatter — proper percent style with sign-strategy, replaces the .percent/.percentSignLess types on CurrencyFormatter.
- NumericFormatter — Double-input formatting with optional unit suffix and locale-aware parsing (was string(double:symbol:) / double(from:) hacks).
- AbbreviatedFormatter — made public, used directly where K/M/B is needed.
- CurrencyFormatter — only fiat currency display (.currency / .fiat / .abbreviated). Internals on Double.FormatStyle.

Shared precision constants, thresholds, and adaptive(for:) selector live in one Precision extension. CurrencyFormatter.precision tuple-switch flattened to a nested switch (no anonymous patterns).

ValueFormatter unchanged vs main except threshold-name aliases (twoPlaces → upToTwoPlaces, fourPlaces → upToFourPlaces) to disambiguate from CurrencyFormatter's pad-zeros twoPlaces.

- [x] Test


gemcoder21 (approved):

ios/Packages/Formatters/Sources/PercentFormatter.swift
```diff
@@ -0,0 +1,25 @@
...
+import Foundation
+
+public struct PercentFormatter: Sendable, Hashable {
+    public static let signed = PercentFormatter(signed: true)
+    public static let unsigned = PercentFormatter(signed: false)
+
+    private let locale: Locale
+    private let signed: Bool
+
+    public init(locale: Locale = .current, signed: Bool = true) {
```
DRadmir: use enum for signed

PR #349, opened 2026-05-19
Round fiat to crypto amount on Android to match iOS

DRadmir:
## Summary

Round fiat → crypto conversion on Send/transfer to display precision so the BigInteger sent in the transaction matches what the user sees.

Before: `$1` in ETH = `0.000473348133351636 ETH` (full divisional precision) shown on Confirm and sent to network.
After: `$1` in ETH = `0.0004733 ETH` (4 significant digits, same as iOS) on both Send equivalent and Confirm header.

Adds `ValueConverter` in `:gemcore` that wraps `ValueFormatter` and applies its precision rules during conversion. `AmountInputType.Fiat.getAmount` and `AmountViewModel.calculateEquivalent` use it.

## Screenshots
Screenshot_20260519_165552
Screenshot_20260519_165556


android/gemcore/src/test/kotlin/com/gemwallet/android/model/ValueConverterTest.kt
```diff
@@ -0,0 +1,37 @@
+package com.gemwallet.android.model
+
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import java.math.BigInteger
+import java.util.Locale
+
+class ValueConverterTest {
```
gemdev111, line 9: can you add few more locales to tests like Locale.GERMANY ?

android/gemcore/src/main/kotlin/com/gemwallet/android/model/ValueConverter.kt
```diff
@@ -0,0 +1,20 @@
+package com.gemwallet.android.model
+
+import com.gemwallet.android.math.parseNumber
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.math.MathContext
+
+class ValueConverter(
```
gemdev111: ValueConverter is stateless - every call site does ValueConverter() . 
Maybe make it simple Kotlin object?

object ValueConverter { }

gemdev111 (approved):

PR #356, opened 2026-05-20
Drop "Balance:" prefix from amount field subtitle (iOS parity, #316)

### LINKED ISSUE #316: Improve amount field subtitle
Hide "Balance:" label
Show just a number: {VALUE} ETH

Image

gemdev111:
iOS shows just the formatted balance value (e.g. "0.02076 ETH") in the AssetBalanceView under the send/stake/perpetual amount input. Android was wrapping the same value with R.string.transfer_balance ("Balance: %s"). Drop the wrapper here; swap keeps the prefix since that surface still uses the Balance: prefix on iOS.

images:

Screenshot_20260520_184428

closes #316


gemcoder21:
Maybe `Balance: ` is a good idea?

gemdev111:
> Maybe `Balance: ` is a good idea?

Added Balance: to iOS instead

PR #357, opened 2026-05-20
Replace custom URL parsing with Gemstone primitives

### LINKED ISSUE #348: Wrong share asset value
On Android, share returns only the contract value instead of the deep link. Consider moving this logic to core.

Image

DRadmir:
## Summary
- Move deeplink + WalletConnect URL parsing/building into the shared Rust core
  (`gemstone`), so iOS and Android use one implementation
- Fix #348: Android asset share now shares a `gemwallet.com` deeplink instead of
  the raw wallet address
- Delete duplicated per-platform parsers (iOS `URLParser`/`DeepLink`, Android
  `WebDeepLinks`/`WalletConnectLinks`/`UriComponentDecoding`)
  
- [x] Test

Closes #348


gemdev111 (approved):

ios/Packages/GemstonePrimitives/Tests/GemstonePrimitivesTests/DeepLinkMock.swift
```diff
@@ -0,0 +1,18 @@
+// Copyright (c). Gem Wallet. All rights reserved.
+
+enum DeepLinkMock {
+    static let assetBitcoin = "https://gemwallet.com/tokens/bitcoin"
```
gemcoder21: does not make sense to have so much code on clients side

ios/Packages/GemstonePrimitives/Sources/URLParser.swift
```diff
@@ -0,0 +1,29 @@
...
+import func Gemstone.walletConnectDecodeUrl
+import enum Gemstone.WalletConnectLink
+import Primitives
+
+enum URLParserError: Error {
+    case invalidURL(URL)
+}
+
+public enum URLParser {
+    public static func from(url: URL) throws -> URLAction {
```
gemcoder21, line 13: we should not have the parser here then? it should be on the gemstone

PR #388, opened 2026-05-27
Improve UTXO sending: Send Max recovery on dust

### LINKED ISSUE #213: Improve UTXO chains sending
User has `0.00237680` on his balance, and he tries to send `0.002376`, it would throw dust error, because `0.00000080` is left, we should calculate this and apply send all in this case?

DRadmir:
Closes #213

When a UTXO transfer (Bitcoin, Litecoin, Doge, etc.) would leave dust as change, the confirm screen now offers a **Max** button that adjusts the amount to send the full balance instead of failing with a dust error.

The pre-existing dust error (the amount itself is below the network's dust threshold, e.g. sending 1 cent) is unchanged and still shows the explanatory info sheet.

Both iOS and Android.

Simulator Screenshot - iPhone Air - 2026-05-27 at 23 08 14
Simulator Screenshot - iPhone Air - 2026-05-27 at 23 08 31
Simulator Screenshot - iPhone Air - 2026-05-27 at 23 08 37


DRadmir:
Fix latter, after migration Bitcoin to core

Issue #392, opened 2026-05-28
Persist WC connection chains locally (Android)


DRadmir:
`BridgesRepository.getConnectionByTopic` fetches chains at runtime from `WalletKit.getListOfActiveSessions()`. If WalletKit is unavailable, we return `null` and the signing request is rejected, even though the connection exists in the DB.

iOS already persists `chains` in `WalletConnectionRecord` and reads them from the DB. Android should mirror this.

**Scope:**
- Add `chains` column to `room_connection` (Room migration).
- Populate from `session.namespaces` in `addConnection`.
- Read from DB in `getConnectionByTopic` instead of the runtime WalletKit call.

Refs: `// TODO Apply chains to local` at [BridgesRepository.kt:121](android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/bridge/BridgesRepository.kt#L121).

PR #393, opened 2026-05-28
Refactor Android WC bridge to use session.topic

### LINKED ISSUE #34: 'val pairingTopic: String' is deprecated. Pairing topic is deprecated.

DRadmir:
## Summary
WalletKit deprecated `Session.pairingTopic`, and our Android bridge was storing it as the connection's session id — architecturally wrong, since pairing topic and session topic are different concepts in WalletConnect v2. This PR switches storage and lookups to `session.topic`.

Local connections are now persisted only after the session is actually established (on `SettledSessionResponse`), instead of right after user approval based on the pairing topic. The owning wallet is resolved from the session's account addresses, ignoring watch-only wallets.

## Notes
- No DB migration. After upgrade, existing entries will be cleaned up by sync on first launch — users will need to re-pair their dApps once.
- Follow-up: persist chains locally — [#392](https://github.com/gemwalletcom/wallet/issues/392).

Closes #34

### DIFF
--- a/android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/bridge/BridgesRepository.kt
+++ b/android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/bridge/BridgesRepository.kt
@@ -8,6 +8,7 @@ import com.gemwallet.android.data.service.store.database.ConnectionsDao
 import com.gemwallet.android.data.service.store.database.entities.DbConnection
 import com.gemwallet.android.data.service.store.database.entities.toDTO
 import com.gemwallet.android.data.service.store.database.entities.toRecord
+import com.gemwallet.android.ext.canSign
 import com.gemwallet.android.ext.walletConnectAppName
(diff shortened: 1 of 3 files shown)


android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/bridge/BridgesRepository.kt
```diff
@@ -317,63 +298,38 @@ class BridgesRepository(
...
-                redirectUniversal = redirect,
-            )
-        )
-    }
-
-    private suspend fun updateSession(session: Wallet.Model.Session) {
-        val room = connectionsDao.getBySessionId(session.pairingTopic) ?: return
-        connectionsDao.update(
-            room.copy(
                 expireAt = System.currentTimeMillis() + session.expiry,
```
gemdev111: can you recheck the exipireAt calulation?

probably should be, compared to iOS
`session.expiry * 1000L`

gemdev111 (approved):

PR #421, opened 2026-06-01
Show unlock date for pending unstaking

DRadmir:
Pending unstaking now shows the unlock date for each request, instead of a single row without a date. Also aligns the staking screens between iOS and Android.
Screenshot 2026-06-01 at 10 35 03 PM
Screenshot 2026-06-01 at 10 56 14 PM

### DIFF
--- a/core/crates/gem_hypercore/src/rpc/client.rs
+++ b/core/crates/gem_hypercore/src/rpc/client.rs
@@ -1,5 +1,5 @@
 use crate::models::{
-    balance::{Balances, DelegationBalance, StakeBalance, Validator},
+    balance::{Balances, DelegationBalance, DelegatorHistoryEntry, StakeBalance, Validator},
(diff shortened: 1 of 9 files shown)


core/crates/gem_hypercore/src/rpc/client.rs
```diff
@@ -115,6 +115,10 @@ impl<C: Client> HyperCoreClient<C> {
         self.info(json!({"type": "delegations", "user": user})).await
     }
 
+    pub async fn get_staking_history(&self, user: &str) -> Result<Vec<DelegatorHistoryEntry>, Box<dyn Error + Send + Sync>> {
```
gemcoder21: there is already `get_delegator_history` method, duplicated, consolidate

gemcoder21 (approved):

PR #448, opened 2026-06-03
cleanup docs for agents

0xh3rman:


DRadmir (approved):
Core has no inner-loop equivalent to the new mobile build-package guidance, should we add one?

0xh3rman:
> Core has no inner-loop equivalent to the new mobile build-package guidance, should we add one?

Agents know Cargo pretty well, if you know only core changes, you can start inside core folder

PR #467, opened 2026-06-08
Add Gem Keystore

0xh3rman:
## Design

Replace Wallet Core's keystore with a Rust-owned **Gem keystore**: the wallet private key stays inside Rust. Signing, message-signing, and device auth all run in core — the apps only call high-level keystore operations, and no raw key crosses the FFI/JNI boundary.

**3 commits, one per platform — please review commit-by-commit** (the combined diff is large). `core` lands first since the apps build on its regenerated UniFFI bindings (bindings are gitignored, so the app commits aren't independently buildable without a stone regen).

### 1. `core` — gem keystore (Rust-owned v4)
- v4 encrypted keystore: Argon2id KDF + AES-256-GCM, deterministic UUIDv5 id, atomic writes, process-global serialization.
- keystore-internal signing, message signing, and device auth (`gem_auth` Ed25519) — the decrypted private key never crosses the FFI boundary; raw-key signers dropped from the UniFFI surface.
- address derived from the public key; recovery-phrase + device-key entropy via `getrandom` (direct OS CSPRNG).
- v3 (Wallet Core) migration reader; Wallet Core removed from core signing, address validation, and fee estimation.
- crypto on the RustCrypto new generation (`aes-gcm 0.11-rc` / `aes 0.9` / `scrypt 0.12` / `getrandom 0.4` / `ripemd 0.2`).

### 2. `android` — migrate to gem keystore
- keystore ops via `withGemKeystore` (decode + zeroize password); transaction signing, message signing, and device auth run in Rust; raw-key `SignClient` / `SignService` removed.
- downgrade-safe v3→v4 migration: v3 file left in place (keyed on v4 presence, no re-run); wallet deletion removes every on-disk copy.
- Trust Wallet Core dependency removed.

### 3. `ios` — migrate to gem keystore
- `LocalKeystore.sign` / `signMessage` / `signAuthMessageHash` via the Rust keystore; old `getPrivateKey` / `ChainSigner` / `SwapSigner` paths removed.
- downgrade-safe v3→v4 migration (v3 left in place); wallet deletion removes every on-disk copy.
- WalletCore SPM package removed.


ios/Packages/FeatureServices/WalletService/WalletService.swift
```diff
@@ -129,6 +128,22 @@ public struct WalletService: Sendable {
         }
     }
 
+    public func migrateV3Keystores() async throws {
+        let wallets = try walletStore.getWallets()
+        for wallet in wallets {
+            do {
+                _ = try await keystore.migrateV3Keystore(for: wallet)
+            } catch {
+                os_log(
```
gemcoder21: use `debugLog`

PR #474, opened 2026-06-09
Single origin check for WalletConnect

### LINKED ISSUE #469: Single origin check for WalletConnect
## Now

The same "trust this session or is it a scam" check is written in three places, three different ways:

- `WCAuthViewModel.kt:71` — returns the status, the caller decides
- `WCRequestViewModel.kt:225` — throws a "scam" error itself
- `ProposalSceneViewModel.kt:81` — decides right inside the screen

This is a security check. Three copies means it's easy to fix one and forget the other two, and let a scam site slip through.

## What to do

Make one place that answers "allow / this is a scam" and use it in all three. Keep the rule — "UNKNOWN and VERIFIED pass, INVALID and MALICIOUS get rejected" — in exactly one spot.

## Don't break

The session proposal screen shows a verification badge — the status must still reach the UI, not just a yes/no.

## Done when

The check lives in one place, all three screens call it, the scam case is still rejected, and the verification badge still works.

Risk: medium (touches security). Size: small.

DRadmir:
The same check for whether a WalletConnect session is trustworthy or a scam was written in three different screens. Now it lives in one place, so the rule can't drift between screens and a scam site can't slip through one of them.

The verification badge on the connection screen still works as before.

Closes #469

### DIFF
--- a/android/features/bridge/viewmodels/src/main/kotlin/com/gemwallet/android/features/bridge/viewmodels/WCRequestViewModel.kt
+++ b/android/features/bridge/viewmodels/src/main/kotlin/com/gemwallet/android/features/bridge/viewmodels/WCRequestViewModel.kt
@@ -11,7 +11,7 @@ import com.gemwallet.android.data.repositories.wallets.WalletsRepository
 import com.gemwallet.android.ext.getAccount
 import com.gemwallet.android.features.bridge.viewmodels.model.BridgeRequestError
 import com.gemwallet.android.features.bridge.viewmodels.model.WCRequest
-import com.gemwallet.android.features.bridge.viewmodels.model.map
+import com.gemwallet.android.features.bridge.viewmodels.model.WalletConnectOriginVerifier
 import com.reown.walletkit.client.Wallet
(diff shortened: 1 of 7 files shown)


gemdev111 (approved):

gemcoder21 (approved):

android/features/bridge/viewmodels/src/main/kotlin/com/gemwallet/android/features/bridge/viewmodels/WCRequestViewModel.kt
```diff
@@ -68,7 +68,9 @@ class WCRequestViewModel @Inject constructor(
                 }
 
                 val appMetadata = connection.session.metadata
-                validateSession(appMetadata.url, verifyContext)
+                if (originVerifier.verify(appMetadata.url, verifyContext).isScam) {
+                    throw BridgeRequestError.ScamSession
```
gemcoder21:
```suggestion
                    throw BridgeRequestError.MaliciousSession
```

gemcoder21 (approved):

PR #495, opened 2026-06-11
Add metadata and tweak deps for fdroid build

0xh3rman:
GitLab MR: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/40300


android/app/src/main/kotlin/com/gemwallet/android/MainViewModel.kt
```diff
@@ -50,7 +50,13 @@ class MainViewModel @Inject constructor(
 
     private val walletConnectHandler = object : PendingNavigationCoordinator.WalletConnectHandler {
         override fun onPairing(uri: String) = addPairing(uri)
-        override fun onRequest() = showWalletConnectPairingToast()
+        override fun onRequest() {
+            if (BuildConfig.WALLET_CONNECT_ENABLED) {
```
gemcoder21: let's avoid having too much code handling WalletConnect, just do return, no need to show any errors.

PR #508, opened 2026-06-15
Migrate deprecated EncryptedSharedPreferences to Tink backed store

### LINKED ISSUE #204: Migrate to new security store based on tink
EncryptedSharedPreferences is deprecated.
We should migrate to new encrypted store. Selected tink + preferences.

0xh3rman:
## Summary
- Replace deprecated secure prefs with a shared Tink-backed store and lazy migration.
- Cover password/Gem migrations and AAD failure with tests.

Fixes #204

## Checks
- `./gradlew :app:assembleGoogleDebug`
- `./gradlew :app:testGoogleDebugUnitTest --tests com.gemwallet.android.data.password.TinkPasswordStoreTest`
- `./gradlew :app:connectedGoogleDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.gemwallet.android.data.password.TinkEncryptedKeyValueStoreInstrumentedTest`


android/app/src/main/kotlin/com/gemwallet/android/data/password/SecureStringStore.kt
```diff
@@ -0,0 +1,19 @@
...
+    fun putString(key: String, value: String)
+
+    fun removeString(key: String): Boolean
+}
+
+internal fun SecureStringStore.getOrMigrate(legacyStore: SecureStringStore, key: String): String? {
+    val value = getString(key) ?: legacyStore.getString(key)?.also {
+        putString(key, it)
+    } ?: return null
+    legacyStore.removeString(key)
```
DRadmir, line 21: This runs on every read, even when the value came from the new store. For upgraded users pwd.xml always exists, so every getPassword/get does a synchronous encrypted commit() to delete a key that's no longer there. Only remove from legacy when a value was actually migrated.

android/app/src/main/kotlin/com/gemwallet/android/data/password/TinkSecurityStore.kt
```diff
@@ -6,49 +6,67 @@ import androidx.datastore.preferences.core.stringPreferencesKey
...
+        config = DEVICE_KEYS_STORE_CONFIG,
+        aeadProvider = aeadProvider,
+    )
+
+    override suspend fun getValue(key: Any): String = withContext(Dispatchers.IO) {
+        val keyValue = key.toString()
+        val value = encryptedStore.getString(keyValue) ?: getLegacyValue(keyValue)?.also {
+            encryptedStore.putString(keyValue, it)
+        } ?: throw IllegalStateException("Data not found")
+        removeLegacyValue(keyValue)
```
DRadmir, line 54: removeLegacyValue runs on every successful read. A transient DataStore write error here throws out of getValue even though we already have the key - and GetDeviceIdImpl catches it and regenerates the device keys, changing the device id. Only remove after an actual migration.

DRadmir (approved):

PR #519, opened 2026-06-16
Fix sending hyper core token

0xh3rman:


gemdev111 (approved):

core/crates/gem_hypercore/src/models/token.rs
```diff
@@ -18,6 +18,12 @@ impl SpotToken {
     }
 }
 
+pub fn spot_token_id_for_asset_id(asset_id: &AssetId) -> Option<String> {
+    let (symbol, contract, _) = asset_id.token_components()?;
```
gemdev111, line 22: `
if asset_id.chain != Chain::HyperCore {
        return None;
    }` or throw?

0xh3rman: This method us only used in hypercore/spot, caller chain is hypercore

PR #549, opened 2026-06-22
Register device before opening the events stream

### LINKED ISSUE #501: Errors: /v2/devices/stream 404
many errors related to this, most likely because app tries to connect to the stream before device is created, make sure it's created before starting connection

DRadmir:
Connect to the device events stream only after the device is registered, so it no longer fails with a 404 on a fresh install. Applies to iOS and Android.

Closes #501

### DIFF
--- a/android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/di/AssetsModule.kt
+++ b/android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/di/AssetsModule.kt
@@ -7,6 +7,8 @@ import com.gemwallet.android.application.fiat.coordinators.SyncFiatTransactions
 import com.gemwallet.android.application.pricealerts.coordinators.UpdatePriceAlerts
 import com.gemwallet.android.blockchain.services.BalancesService
 import com.gemwallet.android.blockchain.services.PerpetualService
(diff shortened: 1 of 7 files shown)


android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/di/AssetsModule.kt
```diff
@@ -148,6 +148,8 @@ object AssetsModule {
         deviceRequestSigner: DeviceRequestSigner,
         streamSubscriptionService: StreamSubscriptionService,
         eventHandler: StreamEventHandler,
+        syncDeviceInfo: com.gemwallet.android.cases.device.SyncDeviceInfo,
```
gemdev111: Add imports
`import com.gemwallet.android.cases.device.IsDeviceRegistered
import com.gemwallet.android.cases.device.SyncDeviceInfo
`

```
// before
        syncDeviceInfo: com.gemwallet.android.cases.device.SyncDeviceInfo,
        isDeviceRegistered: com.gemwallet.android.cases.device.IsDeviceRegistered,
// after
        syncDeviceInfo: SyncDeviceInfo,
        isDeviceRegistered: IsDeviceRegistered,

```

gemdev111 (approved):

PR #553, opened 2026-06-23
Fix TRON address transaction history 500 on missing receipt

0xh3rman:
## Problem

`GET /v1/chain/address/tron/{address}/transactions` returned **HTTP 500** (`missing field \`id\``) whenever any transaction's receipt was missing. Users saw incoming TRX transfers as missing (outgoing still showed because the app sources sent txs from its local DB).

Root cause: TRON's `gettransactioninfobyid` returns an empty object `{}` (HTTP 200) for a transaction the node has not indexed yet (recent tx / fullnode lag). `{}` fails to deserialize into `TransactionReceiptData`, and `try_join_all` propagated that single failure, tanking the whole address request.

## Fix

`get_transaction_reciept` now returns `Option<TransactionReceiptData>`:
- empty `{}` body → `None` (not indexed yet)
- any other body — malformed JSON or a transport error — still **propagates** (no silent data loss)

Callers handle the typed missing case:
- **`get_transactions_by_address`** keeps `try_join_all` (errors propagate) and drops only the transactions whose receipt is `None`.
- **`get_transaction_by_hash`** returns `Ok(None)` for a missing receipt.
- **`get_transaction_status`** maps a missing receipt to `Pending` (consistent with the existing `block_number == 0` branch).

This fixes the `{}`-induced 500 on all three TRON receipt paths.

## Tests

- Unit test via `mock_tron_client` (new testkit helper): a `{}` receipt is dropped while the rest map; malformed and transport errors fail the whole request rather than silently dropping data. Fixtures (`transactions_by_address.json`, `transactions_by_address_receipts.json`) are real responses captured for `TBKwjUtXVsX1r724C1V52nocBgtioDjx9u`.
- Integration `test_get_transactions_by_address` bumped from limit 1 → 2 so it no longer depends on the single most-recent tx being a mapper-supported type.
- Verified live: `get_transaction_by_hash`, `get_transactions_by_block`, and a by-address fetch for another address all pass.

### DIFF
--- a/core/crates/gem_tron/src/provider/transactions.rs
+++ b/core/crates/gem_tron/src/provider/transactions.rs
@@ -21,11 +21,10 @@ impl<C: Client + Clone> ChainTransactions for TronClient<C> {
     }
 
     async fn get_transaction_by_hash(&self, hash: String) -> Result<Option<Transaction>, Box<dyn Error + Sync + Send>> {
-        Ok(map_transaction(
-            self.get_chain(),
-            self.get_transaction(hash.clone()).await?,
-            self.get_transaction_reciept(hash).await?,
-        ))
+        let Some(receipt) = self.get_transaction_reciept(hash.clone()).await? else {
+            return Ok(None);
+        };
+        Ok(map_transaction(self.get_chain(), self.get_transaction(hash).await?, receipt))
     }
 
     async fn get_transactions_by_address(&self, request: TransactionsRequest) -> Result<Vec<Transaction>, Box<dyn Error + Sync + Send>> {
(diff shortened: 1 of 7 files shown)


core/crates/gem_tron/src/provider/transactions.rs
```diff
@@ -39,11 +38,80 @@ impl<C: Client + Clone> ChainTransactions for TronClient<C> {
...
+    use crate::provider::testkit::mock_tron_client;
+    use gem_client::ClientError;
+
+    const TRANSACTIONS_RESPONSE: &str = include_str!("../../testdata/transactions_by_address.json");
+    const RECEIPTS_RESPONSE: &str = include_str!("../../testdata/transactions_by_address_receipts.json");
+    const ADDRESS: &str = "TBKwjUtXVsX1r724C1V52nocBgtioDjx9u";
+    const LAGGING_TRANSACTION_ID: &str = "e5c5dc535b267134024e887c00b1522426661b1b5ae6efb76606f4d83bca1a81";
+    const INCOMING_TRANSACTION_ID: &str = "7b633cd06802d7117a7202650c7580516c742ce1e20d43ba736ab8da1a02cd8f";
+
+    async fn mock_fetch(receipt_handler: impl Fn(&str) -> Result<Vec<u8>, ClientError> + Send + Sync + 'static) -> Result<Vec<Transaction>, Box<dyn Error + Sync + Send>> {
```
gemcoder21, line 63: useless mock, remove it, this creates bad patterns to follow

0xh3rman: should be removed from main now

PR #573, opened 2026-06-25
Deduplicate add-missing-balance logic

DRadmir:
Unifies three duplicated "add balance row if missing" code paths into a single `BalanceStore.addBalance` overload that relies on the existing insert-or-ignore. No behavior change.


ios/Packages/FeatureServices/BalanceService/BalanceService.swift
```diff
@@ -86,29 +86,11 @@ extension BalanceService {
     }
 
     public func addAssetsBalancesIfMissing(assetIds: [AssetId], wallet: Wallet, isEnabled: Bool?) throws {
```
gemcoder21:
```suggestion
    public func addAssetsBalancesIfMissing(assetIds: [AssetId], wallet: Wallet, isEnabled: Bool) throws {
```

Issue #570, opened 2026-06-24
Split BridgesRepository: extract ConnectionsRepository from WalletConnect orchestration


gemcoder21:
## Summary

`BridgesRepository` (`android/data/repositories/.../bridge/BridgesRepository.kt`, ~343 lines) mixes two distinct responsibilities. Now that the F-Droid WalletConnect refactor has introduced a clean `WalletConnectClient` abstraction, the repository is a good candidate to split along its natural seam in a separate, independently-reviewable PR.

This is a follow-up split off from the F-Droid WalletConnect isolation work — intentionally **not** done in that PR because it touches signing/auth flows and deserves its own review.

## Current responsibilities

1. **Connection persistence (Room)** — `getConnections`, `getConnection(ById/ByTopic)`, `addConnection`, `updateConnection`, `deleteConnection`, `sync`, `disconnect`, `toConnectionRecord`, `getSupportedNamespaces`, `walletForSession`. Depends only on `ConnectionsDao` + `WalletsRepository`.
2. **Protocol orchestration** — `init {}`, `bridgeEvents`, `initWalletConnect`, `addPairing`, `approveConnection`, `handlePendingRequests`, `pingActiveSessions`. Genuinely combines the client *and* the DB.
3. **Thin client passthroughs** — `rejectConnection`, `approveAuthentication`, `rejectAuthentication`, `respondSessionRequest`, `generateAuthPayloadParams`, `formatAuthMessage`, `generateAuthObject`.

## Proposed change

Extract responsibility **#1** into a `ConnectionsRepository` (Room CRUD + mapping). `BridgesRepository` then becomes a pure orchestrator depending on `ConnectionsRepository` + `WalletConnectClient`, dropping to ~180 lines.

Keep responsibility **#3 (passthroughs) in `BridgesRepository`** — they exist deliberately so feature view models never import `WalletConnectClient` directly. That decoupling is the point of the abstraction; do not push it back into the feature modules.

## Out of scope / non-goals

- Do not re-expose `WalletConnectClient` to feature modules.
- No behavior change to signing, auth, or session approval flows — pure structural refactor.

## Acceptance criteria

- [ ] Room persistence + mapping lives in a `ConnectionsRepository`.
- [ ] `BridgesRepository` orchestration delegates persistence to `ConnectionsRepository`.
- [ ] Feature view models still depend only on `BridgesRepository` (no `WalletConnectClient` / `com.reown` imports).
[...]

PR #579, opened 2026-06-26
Keep WalletConnect connections when active-session fetch fails

DRadmir:
When the WalletConnect SDK failed to return the active session list, BridgesRepository treated the empty result as "no sessions" and deleted every local connection. It now returns null on failure and skips the reconciliation, so connections are kept. Includes unit tests for the null and empty cases.

Draft.

### DIFF
--- a/android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/bridge/BridgesRepository.kt
+++ b/android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/bridge/BridgesRepository.kt
@@ -1,5 +1,6 @@
 package com.gemwallet.android.data.repositories.bridge
 
+import android.util.Log
 import androidx.core.net.toUri
 import com.wallet.core.primitives.WalletConnection
 import com.wallet.core.primitives.Wallet as GemWallet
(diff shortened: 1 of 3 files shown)


gemcoder21 (approved):

android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/bridge/BridgesRepository.kt
```diff
@@ -88,26 +89,27 @@ class BridgesRepository(
...
+    private fun activeSessions(): List<WalletConnectSession>? =
         runCatching { walletConnectClient.activeSessions().filter { it.metadata != null } }
-            .getOrDefault(emptyList())
+            .onFailure { Log.e("BridgesRepository", "Failed to get active sessions", it) }
+            .getOrNull()
 
     suspend fun disconnect(id: String, onSuccess: () -> Unit = {}, onError: (String) -> Unit = {}) {
         val connection = connectionsRepository.disconnect(id) ?: return
-        val activeSession = activeSessions().firstOrNull { it.topic == connection.session.sessionId }
+        val activeSession = activeSessions()?.firstOrNull { it.topic == connection.session.sessionId }
```
gemcoder21, line 112: looks like a lot of duplicated code, can you review this file? or maybe even move, see https://github.com/gemwalletcom/wallet/issues/570

PR #580, opened 2026-06-26
Harden keystore secret handling and durability

0xh3rman:
Zeroize transient private-key decode buffers (the discarded tail of a base58 keypair, the base32 buffer, and failed-validation/hex temporaries) so they are wiped on drop instead of left in freed heap.

Return the sanitized mnemonic as Zeroizing<String> and wipe the normalize/lowercase intermediates, so phrase-validation paths leave no plaintext copy behind.

fsync the keystore directory after deleting a secret file, so a crash cannot resurrect a deleted secret (mirrors the write path).

Recover from a poisoned keystore queue lock (the guarded value is ()) instead of permanently failing all keystore operations after an unrelated panic.

### DIFF
--- a/core/crates/gem_keystore/src/mnemonic.rs
+++ b/core/crates/gem_keystore/src/mnemonic.rs
@@ -14,15 +14,15 @@ impl Mnemonic {
         Ok(mnemonic.words().map(|word| word.to_string()).collect())
     }
 
-    pub fn sanitize(phrase: &str) -> Result<Zeroizing<String>, KeystoreError> {
-        let sanitized = Zeroizing::new(phrase.nfkd().collect::<String>());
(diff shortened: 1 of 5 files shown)


core/crates/gem_keystore/src/mnemonic.rs
```diff
@@ -14,11 +14,11 @@ impl Mnemonic {
         Ok(mnemonic.words().map(|word| word.to_string()).collect())
     }
 
-    pub fn sanitize(phrase: &str) -> Result<String, KeystoreError> {
-        let sanitized = phrase.nfkd().collect::<String>();
-        let sanitized = sanitized.split_whitespace().map(|word| word.to_lowercase()).collect::<Vec<_>>().join(" ");
+    pub fn sanitize(phrase: &str) -> Result<Zeroizing<String>, KeystoreError> {
```
gemcoder21: can we avoid `sanitize` does not really explain what it does

gemcoder21 (approved):

PR #592, opened 2026-07-01
Preflight device registration and wallet subscription for REST calls

### LINKED ISSUE #550: Android: device REST calls can 404 before device is registered
On a fresh install, device-authed REST calls (e.g. `/v2/devices/fiat/assets/buy`) can fire before the device is registered and return 404 "Device not found".

iOS already waits for registration before such calls; Android has no equivalent. Add the same registration preflight on Android.

Related: #501 (the WebSocket side of the same race).

### LINKED ISSUE #603: Android: new wallet getAssets 404 "Wallet not found"
On a freshly created wallet, `GET /v2/devices/assets` can return 404 "Wallet not found": the wallet-scoped asset fetch fires before the wallet is subscribed on the backend.

## Cause
`PhraseAddressImportWalletService.createWallet` activates the session before subscribing the wallet:

1. `setupWallet(wallet)` → `sessionRepository.setWallet(wallet)` — activating the wallet eagerly wakes `StreamObserverService`'s session collector → `syncAssets()` → `DeviceAssetsSyncService.sync` → `GET /v2/devices/assets` (needs the wallet subscribed).
2. Only afterwards `syncSubscription.syncSubscription(...)` runs → `POST /v2/devices/subscriptions`.

The two run on separate coroutines with no ordering, so `getAssets` can win the race → backend doesn't know the wallet yet → 404.

## Evidence (Proxyman, fresh install + create wallet)
- `GET /v2/devices/assets?from_timestamp=0` → 404 `{"error":{"message":"Wallet not found device_id=… wallet_id=multicoin_0x…"}}`
- fires before `POST /v2/devices/subscriptions` → 200

Device is registered (device_id present) — this is NOT the device-registration race, it's the wallet-subscription precondition.

## Fix direction
Subscribe the wallet before activating the session (analogous to the device-registration fix):

```kotlin
assetsRepository.createAssets(wallet)
runCatching { syncSubscription.syncSubscription(listOf(wallet)) }  // before setWallet
sessionRepository.setWallet(wallet)
```

Apply the same to `importWallet` (check `walletImportSync.sync` isn't too heavy to move before activation).

## Related
Same class as the device-registration (#550) and stream (#501) races, one level up. Android only.

### LINKED ISSUE #620: Empty Buy assets after first wallet creation
Image

DRadmir:
Wallet-scoped device REST calls could 404 on fresh state:
- **Device not found** — the call fired before the device was registered.
- **Wallet not found** — a new wallet's assets were fetched before it was subscribed on the backend.

Both are fixed with a transport-level preflight: wallet-tagged requests wait until the device is registered and its subscriptions are synced before proceeding. Subscriptions are marked dirty on wallet create/import and reconciled lazily, matching iOS.

Closes #550
Closes #603
Closes https://github.com/gemwalletcom/wallet/issues/620

## TODO: testing
- [x] Manual on-device testing

### DIFF
--- a/android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/device/DeviceRepository.kt
+++ b/android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/device/DeviceRepository.kt
@@ -67,6 +67,7 @@ class DeviceRepository(
     private val Context.dataStore by preferencesDataStore(name = "device_config")
 
     private val registrationMutex = Mutex()
+    private val subscriptionMutex = Mutex()
 
     override suspend fun syncDeviceInfo() {
         synchronizeDevice(
@@ -158,6 +159,11 @@ class DeviceRepository(
 
     override suspend fun invoke() {
         if (isDeviceRegistered()) return
+        syncDeviceInfo()
+    }
+
+    suspend fun ensureSubscriptionsSynced() = subscriptionMutex.withLock {
+        if (!hasPendingSubscriptionChanges()) return@withLock
         syncDeviceInfo()
     }
 
@@ -299,7 +305,7 @@ class DeviceRepository(
         configStore.putBoolean(Keys.SubscriptionVersionHasChange.string, hasChange)
     }
 
-    private fun invalidateSubscriptions() {
+    fun invalidateSubscriptions() {
         val invalidatedState = getSubscriptionSyncState().invalidate()
         setSubscriptionVersion(invalidatedState.version)
         setSubscriptionVersionHasChange(invalidatedState.hasPendingChanges)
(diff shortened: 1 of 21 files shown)


android/data/repositories/src/main/kotlin/com/gemwallet/android/data/repositories/device/DeviceRepository.kt
```diff
@@ -150,18 +156,23 @@ class DeviceRepository(
         return context.dataStore.data.map { it[Key.DeviceRegistered] }.firstOrNull() == true
     }
 
-    private suspend fun getOrCreateDevice(device: Device): Device {
+    override suspend fun invoke() {
```
gemdev111, line 164: Can you recheck this please:

Wallet calls that fire during the first registration each also run a full `syncDeviceInfo()` - they pass the `isDeviceRegistered()` check before the flag flips. The mutex prevents double-register, but each still does an extra `GET /devices`. A few redundant calls on first launch

iOS coalesces this in its sync coordinator

gemdev111 (approved):

PR #658, opened 2026-07-09
Add OKX Tron swap support and balance diff for tron swap tx

0xh3rman:
Simulator Screenshot - iPhone 17 Pro - 2026-07-07 at 22 17 08


core/crates/gem_tron/src/provider/transactions_mapper.rs
```diff
@@ -60,6 +86,34 @@ pub fn map_transaction(chain: Chain, transaction: TronTransaction, receipt: Tran
         let from = contract_value.owner_address.unwrap_or_default();
 
         let contract_type = value.contract_type;
+        let logs = receipt.log.unwrap_or_default();
+
+        if contract_type == Some(TronContractType::TriggerSmart)
+            && let Some(owner) = TronAddress::from_hex_or_base58(&from)
+            && let Some(swap) = tron_swap_metadata(chain, &owner, contract_value.call_value, &logs)
```
DRadmir: For USDT - TRX, the received TRX comes via an internal tx with no ERC20 log and call_value == 0, so there's only one non-zero delta and map_swap returns None. The swap isn't detected, and if logs.len() > 1 it also drops from history via the single-log handler below. Intended, or should we handle it?

core/crates/gem_tron/src/provider/transactions_mapper.rs
```diff
@@ -18,6 +21,29 @@ fn resource_type_metadata(resource: Option<String>) -> Option<serde_json::Value>
...
+    if let Some(call_value) = call_value.filter(|value| *value > 0) {
+        balance_diffs.push(BalanceDiff {
+            asset_id: chain.as_asset_id(),
+            from_value: None,
+            to_value: None,
+            diff: -BigInt::from(call_value),
+        });
+    }
+
+    SwapMapper::map_swap(&balance_diffs, &BigUint::from(0u8), &chain.as_asset_id(), None)
```
DRadmir, line 64: Passing provider: None here means tron swaps won't show a provider name in history. Intentional because it can't be derived from logs?

DRadmir (approved):

core/crates/swapper/src/okx/mod.rs
```diff
@@ -4,6 +4,36 @@ mod constants;
...
+#[cfg(test)]
+mod testkit;
+
+use crate::SwapperChainAsset;
+use primitives::Chain;
 
 pub use model::OkxClientConfig;
 pub use provider::OkxProvider;
+
+pub(crate) fn support_assets() -> Vec<SwapperChainAsset> {
```
gemcoder21: should this be inside swapper okx dex provider?

core/crates/swapper/src/okx/provider.rs
```diff
@@ -110,6 +111,25 @@ where
         Ok(SwapQuoteData::new_contract(tx.to.clone(), value, tx.data.clone(), approval, gas_limit))
     }
 
+    async fn build_tron_quote_data(&self, tx: &TransactionData, from_asset: &QuoteAsset, from_value: &str, owner: &str) -> Result<SwapQuoteData, SwapperError> {
```
gemcoder21:
```suggestion
    async fn build_tron_quote_data(&self, transaction_data: &TransactionData, from_asset: &QuoteAsset, from_value: &str, owner: &str) -> Result<SwapQuoteData, SwapperError> {
```

core/crates/swapper/src/okx/provider.rs
```diff
@@ -328,34 +371,22 @@ mod tests {
...
     #[test]
     fn test_build_solana_quote_data() {
-        let tx = TransactionData {
-            data: "Cn8eVZg".to_string(),
-            to: "ToAddr".to_string(),
-            value: String::new(),
-            gas: String::new(),
-            signature_data: None,
-        };
+        let tx = TransactionData::mock("ToAddr", "", "Cn8eVZg", "");
```
gemcoder21:
```suggestion
        let transaction = TransactionData::mock("ToAddr", "", "Cn8eVZg", "");
```

gemcoder21 (approved):

core/crates/gem_tron/src/provider/transactions_mapper.rs
0xh3rman: Yes, we're using generic diff method instead of registry based for TRON too

core/crates/gem_tron/src/provider/transactions_mapper.rs
0xh3rman: it turns out tron returns internal tx so should be fixed

PR #725, opened 2026-07-17
Throttle transaction status polling to provider rate limit

### LINKED ISSUE #716: Swap status polling hits the provider's rate limit
The app polls swap status faster than the provider allows. Around half of the requests are rejected on iOS, about 8% on Android.

```
GET /proxy/swap/near-intents/explorer/api/v0/transactions?search=<hash>
→ {"code":"RATE_LIMIT_EXCEEDED","message":"Rate limit exceeded. Please wait 1 second(s) before making another request."}
```

The upstream allows ~1 request/sec and there is no backoff — 429 and 200 alternate. Status still resolves via the polls that get through, so it is not user-breaking today, but a tighter upstream limit would stall swap-status tracking.

Throttle to the documented limit and add backoff on 429.

DRadmir:
The app polled swap status faster than the provider allows, so a large share of the requests came back rejected.

The interval was taken from the chain's block time, which on fast chains meant two requests a second. It now never starts faster than once per second. On iOS the first check also fired the instant a transaction was sent, when the block could not have confirmed yet — that request is gone.

Closes #716

### DIFF
--- a/core/crates/primitives/src/job_configuration.rs
+++ b/core/crates/primitives/src/job_configuration.rs
@@ -21,7 +21,7 @@ impl JobConfiguration {
     pub fn transaction_state(chain: Chain) -> Self {
         let config = Self::default();
         Self {
-            initial_interval_ms: chain.block_time().clamp(1_000, config.initial_interval_ms),
+            initial_interval_ms: chain.block_time().clamp(1, config.initial_interval_ms),
             ..config
         }
     }
@@ -57,8 +57,6 @@ mod tests {
         assert_eq!(config.initial_interval_ms, 5_000);
         assert_eq!(config.max_interval_ms, 60_000);
         assert_eq!(config.step_factor, 1.1);
-
-        assert_eq!(JobConfiguration::transaction_state(Chain::Solana).initial_interval_ms, 1_000);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


core/crates/primitives/src/job_configuration.rs
```diff
@@ -57,6 +57,8 @@ mod tests {
         assert_eq!(config.initial_interval_ms, 5_000);
         assert_eq!(config.max_interval_ms, 30_000);
```
gemcoder21: let's increase internal for 60s?

ios/Packages/Primitives/Sources/JobRunner/JobRunner.swift
```diff
@@ -49,10 +53,6 @@ extension JobRunner {
                 }
                 return
             case let .retry(error):
-                let sleepUntil = attemptStart.advanced(by: .milliseconds(Int(intervalMs)))
```
gemcoder21, line 52: why was this removed?

gemcoder21 (approved):

ios/Packages/Primitives/Sources/JobRunner/JobRunner.swift
DRadmir: The loop now sleeps before each request instead of after, so this offset isn’t needed anymore. It also removes the first request that was sent right after transaction was sent.

PR #728, opened 2026-07-17
Fix fiat input producing a 1000x too small amount

### LINKED ISSUE #719: Fiat amount input converts 1000× too low on whole amounts
In fiat input mode, entering a round amount produces a crypto amount 1000× too small, and Continue accepts it.

Reproduced on `origin/main` with a live USDT price: enter **$1000** with a balance of 5.81 USDT → the app computes **1 USDT** (should be 1000) and enables Continue, because 1 fits the balance.

Root cause is a format→parse round-trip in `ValueConverter.convertToAmount`:

```kotlin
val display = formatter.string(Crypto(amount, decimals).atomicValue, decimals)  // "1,000"
return Crypto(display.parseNumber(), decimals)                                  // 1.000
```

`ValueFormatter` builds on `NumberFormat.getInstance(locale)` (`ValueFormatter.kt:36`), which has grouping enabled by default, and `Precision` (`Precision.kt:21-29`) never disables it. So the app formats `1000` into `"1,000"` and immediately re-parses its own display string.

`parseNumber` (`NumberParser.kt:6`) places the decimal point before the last group. That is correct when a fractional part exists — `"1,000.5"` → `1000.5`, `"1,5"` → `1.5` — but with grouping and no fraction it reads the last group as decimals: `"1,000"` → `1.000`.

This feeds the real send amount: `AmountInputType.Fiat.getAmount` → `AmountViewModel.onNext:147` → `buildConfirmParams`.

Conditions: fiat mode, result ≥ 1000, whole after rounding to 2 places. Stablecoins hit it directly, since price is exactly 1.0. `2000.5` is unaffected — the fractional part masks it. `ValueConverterTest` only uses small values (`1/76800`), so tests never see grouping.

Fix: stop round-tripping through the display string — round the `BigDecimal` directly.

Image

DRadmir:
Entering **$1000** in fiat mode sent **1 USDT**, and Continue accepted it because 1 fit the balance.

The amount was turned into text for the screen and then read back out of that text. A grouped number like `1,000` came back as `1`. Now it stays a number and is never rendered and re-read.

iOS had the same bug, in locales that group thousands with a dot — fixed here too.

Closes #719


ios/Packages/Formatters/Tests/FormattersTests/NumberInputNormalizerTests.swift
```diff
@@ -177,6 +177,27 @@ struct NumberInputNormalizerTests {
...
+            (Locale.PT_BR, "12.345.678", "12345678"),
+            (Locale.US, "1.000.000", "1000000"),
+            (Locale.DA_DK, "1,000,000", "1000000"),
+            (Locale.DA_DK, "1.234", "1234"),
+            (Locale.DA_DK, "1,5", "1.5"),
+            (Locale.US, "1.5", "1.5"),
+            (Locale.US, "1.234", "1.234"),
+        ]
+        for testCase in testCases {
+            let normalized = NumberInputNormalizer.normalize(testCase.input, locale: testCase.locale)
```
gemcoder21: can we never use word `normalize` in the codebase.

ios/Packages/Formatters/Sources/NumberInputNormalizer.swift
```diff
@@ -53,25 +53,28 @@ enum NumberInputNormalizer {
...
                 result = result.replacingOccurrences(of: comma, with: String.empty)
+            } else {
+                result = result.replacingOccurrences(of: comma, with: dot)
             }
         }
 
         return result
     }
 
+    private static func occursMoreThanOnce(_ symbol: String, in s: String) -> Bool {
```
gemcoder21: maybe add extension to String? to check `occurrences` and then in `if occurrences > 1`

gemcoder21 (approved):

PR #739, opened 2026-07-20
Skip Sui staking reward simulations when there are no staked objects

gemdev111:
create_delegated_stakes ran two SimulateTransaction calls even with no staked objects, on every Sui balance refresh. Return early on empty and join the two simulations for wallets that do stake.


gemcoder21:
@0xh3rman can you review this one?

0xh3rman:
Thanks for the fix, lgtm

PR #748, opened 2026-07-21
Clear transactions and preferences when deleting a wallet

### LINKED ISSUE #711: Android: deleting a wallet leaves its data behind
Delete a wallet and its transaction history and settings stay in the app's database. Importing the same phrase again picks them back up.

iOS removes everything.

Found during regression testing on a clean install (import → delete → re-import). After delete: wallet gone, but 100 transaction rows and `wallet_preferences_<walletId>.xml` remain. Accounts, addresses, balances, session, NFT associations and key material are all cleared correctly.

DRadmir:
Deleting a wallet on Android left its transaction history and settings in the app. Importing the same phrase brought them back. iOS already removes everything.

Now deleting a wallet also clears its transactions and preferences, matching iOS.

Closes #711


android/data/services/store/src/main/kotlin/com/gemwallet/android/data/service/store/database/di/Migration_82_83.kt
```diff
@@ -0,0 +1,46 @@
...
+                "`fee` TEXT NOT NULL, " +
+                "`value` TEXT NOT NULL, " +
+                "`payload` TEXT, " +
+                "`direction` TEXT NOT NULL, " +
+                "`createdAt` INTEGER NOT NULL, " +
+                "`updatedAt` INTEGER NOT NULL, " +
+                "PRIMARY KEY(`id`, `walletId`), " +
+                "FOREIGN KEY(`walletId`) REFERENCES `wallets`(`id`) ON UPDATE CASCADE ON DELETE CASCADE)"
+        )
+        db.execSQL(
```
gemcoder21, line 32: what if wallets already don't exist and it will fail fk?

gemcoder21 (approved):
but I think it will fail, because if any transaction does not have fk to the wallet it would fail

DRadmir:
copy step only carries over transactions whose wallet still exists

PR #751, opened 2026-07-21
Calculate price change and wallet total in core

gemdev111:
Price change math and the wallet total (value, PnL, PnL %) were each implemented twice — iOS and Android had their own copies of the same rules. Both now live in core and the apps use them through gemstone.

- core: `PriceChangeCalculator` (percentage, amount) and `BalanceCalculator.totalFiatValue` in primitives with the math tests; `TotalFiatValue` is now a shared typeshare model
- iOS: thin wrappers in GemstonePrimitives with gem mappers; the Store calculator and the manual `TotalFiatValue` type are removed
- Android: `PriceChangeCalculator` delegates to gemstone; the wallet summary fold and the percentage helper are replaced by the core calculator (totals sum as Double now, same as iOS)
- platform tests keep only formatting and store-pipeline coverage; math is tested in Rust


ios/Packages/Store/Package.swift
```diff
@@ -17,13 +17,15 @@ let package = Package(
...
         .package(name: "Primitives", path: "../Primitives"),
+        .package(name: "GemstonePrimitives", path: "../GemstonePrimitives"),
         .package(name: "GRDB", path: "../../Submodules/GRDB"),
     ],
     targets: [
         .target(
             name: "Store",
             dependencies: [
                 "Primitives",
+                "GemstonePrimitives",
```
gemcoder21: `GemstonePrimitives` should not be imported into Store

DRadmir (approved):

PR #752, opened 2026-07-21
Move autoclose trigger validation to core

gemdev111:
The take profit / stop loss trigger rule (price must be above or below market depending on position direction) was implemented twice. It now lives in core and both apps validate through gemstone.

- core: `AutocloseValidator` returns a closed `AutocloseValidation` verdict (valid, invalid amount, trigger must be higher/lower)
- iOS: validator keeps text parsing and error messages; the rule comes from core
- Android: validator keeps its nullable API and maps the core verdict
- rule tests live in Rust, including the price-equals-market boundary case


android/gemcore/src/main/kotlin/com/gemwallet/android/domains/perpetual/autoclose/AutocloseField.kt
```diff
@@ -1,16 +1,17 @@
...
     val type: TpslType,
     val price: Double?,
     val originalPrice: Double?,
     val formattedPrice: String?,
-    val error: AutocloseError?,
+    val validation: AutocloseValidation,
     val orderId: ULong?,
 ) {
-    val isValid: Boolean get() = price != null && error == null
+    val isValid: Boolean get() = price != null && validation == AutocloseValidation.VALID
```
DRadmir, line 14: should we use uppercase `VALID`?

DRadmir (approved):

android/gemcore/src/main/kotlin/com/gemwallet/android/domains/perpetual/autoclose/AutocloseField.kt
gemdev111: It's uniffi codegen, not a choice here - Rust declares `Valid` / `TriggerMustBeHigher` and the Kotlin emits `VALID` / `TRIGGER_MUST_BE_HIGHER` (Kotlin's enum-constant convention). Typeshare enums like `TpslType.TakeProfit` stay PascalCase for the same reason - each generator follows its target language's style.

PR #753, opened 2026-07-21
Convert between crypto and fiat in core

gemdev111:
Crypto to fiat conversion was implemented twice with a behavior difference: converting fiat to crypto with a zero or negative price silently returned zero on Android, while iOS rejected it. The conversion now lives in core and always fails closed on invalid prices.

- core: `CryptoFiatConverter` in number_formatter — atomic value × price and fiat ÷ price with an explicit error on price ≤ 0; the atomic value crosses the boundary as `BigInt`, so invalid input is rejected before the math runs
- iOS: `ValueConverter` delegates the division; API and behavior unchanged
- Android: `toCrypto` is now nullable — call sites handle the invalid-price case explicitly instead of getting a silent zero
- currency formatting stays on each platform; conversion math is tested in Rust


core/crates/number_formatter/src/crypto_fiat_converter.rs
```diff
@@ -0,0 +1,62 @@
+use bigdecimal::BigDecimal;
+use std::str::FromStr;
+
+use crate::big_number_formatter::{BigNumberFormatter, NumberFormatterError};
+
+pub struct CryptoFiatConverter {}
+
+impl CryptoFiatConverter {
+    pub fn convert_to_fiat(value: &str, decimals: u32, price: f64) -> Result<String, NumberFormatterError> {
```
DRadmir:
```suggestion
    pub fn to_fiat(value: &str, decimals: u32, price: f64) -> Result<String, NumberFormatterError> {
```

core/crates/number_formatter/src/crypto_fiat_converter.rs
```diff
@@ -0,0 +1,62 @@
...
+        let amount = BigNumberFormatter::big_decimal_value(value, decimals)?;
+        Ok((amount * Self::price_value(price)?).normalized().to_string())
+    }
+
+    pub fn convert_to_crypto(fiat_amount: &str, decimals: u32, price: f64) -> Result<String, NumberFormatterError> {
```
DRadmir:
```suggestion
    pub fn to_crypto(fiat_amount: &str, decimals: u32, price: f64) -> Result<String, NumberFormatterError> {
```

DRadmir (approved):

PR #758, opened 2026-07-22
Show indicator dot on swap settings when slippage is custom

DRadmir:
Adds a small blue dot on the Swap settings icon when the slippage is set to a custom value. This way it is easy to notice that slippage was changed and not left at the default.

Simulator Screenshot - wt1 - 2026-07-22 at 14 49 16


ios/Features/Swap/Sources/Scenes/SwapScene.swift
```diff
@@ -55,8 +55,20 @@ public struct SwapScene: View {
         .navigationTitle(model.title)
         .toolbar {
             ToolbarItem(placement: .topBarTrailing) {
-                Button("", systemImage: SystemImage.settings) {
+                Button {
                     isPresentingSlippage = true
+                } label: {
+                    Image(systemName: SystemImage.settings)
+                }
+                .overlay(alignment: .topTrailing) {
```
gemdev111: Lets make component from button (with blue dot indicator) ?

gemdev111 (approved):

PR #761, opened 2026-07-22
Add retry when building Tink Aead

0xh3rman:
## Summary

Fixes two Android Keystore failure modes that can occur while Tink builds the AEAD used by secure storage:

1. **Transient Keystore failures** — retry `Aead` creation up to 3 attempts with linear backoff for `GeneralSecurityException` and `ProviderException`, including the reported `InvalidKeyException` path. Tink 1.22 does not retry when the Keystore master key temporarily fails to load.
2. **Invalid encrypted keysets** — treat `BadPaddingException` / `AEADBadTagException` as non-retryable authentication failures. When a store explicitly opts in, clear its unreadable encrypted values and keyset, then build a fresh keyset.

Invalid-keyset reset is enabled only for `TinkSecurityStore`, where the device signing identity can be regenerated. `TinkPasswordStore` and `TinkGemPreferences` remain fail-closed so this recovery path cannot erase wallet password data or other non-regenerable values.

Request signing now also runs inside the `SecurityInterceptor` error boundary, so a failure that cannot be recovered returns a retryable HTTP 503 instead of escaping on the OkHttp dispatcher thread and crashing the app.

## Impact

- Transient Keystore failures can recover without deleting data.
- A corrupted or mismatched device keyset can self-heal by regenerating the device security store.
- Sensitive stores are never automatically reset.
- Unrecoverable signing failures degrade to HTTP 503 instead of a process crash.

## Tests

- `./gradlew :app:testGoogleDebugUnitTest --tests com.gemwallet.android.data.password.TinkAeadRetryTest`
- `./gradlew :data:services:remote-gem:testDebugUnitTest --tests com.gemwallet.android.data.services.gemapi.http.SecurityInterceptorTest`
- `./gradlew :app:connectedGoogleDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.gemwallet.android.data.password.TinkEncryptedKeyValueStoreInstrumentedTest`
- `./gradlew :app:assembleGoogleDebugAndroidTest`
- `./gradlew :app:lintGoogleDebug`

### DIFF
--- a/android/app/src/androidTest/kotlin/com/gemwallet/android/data/password/TinkEncryptedKeyValueStoreInstrumentedTest.kt
+++ b/android/app/src/androidTest/kotlin/com/gemwallet/android/data/password/TinkEncryptedKeyValueStoreInstrumentedTest.kt
@@ -7,6 +7,10 @@ import androidx.security.crypto.EncryptedSharedPreferences
 import androidx.security.crypto.MasterKey
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.gemwallet.android.math.fromHex
+import com.gemwallet.android.math.hex
+import com.google.crypto.tink.proto.EncryptedKeyset
+import com.google.crypto.tink.shaded.protobuf.ByteString
 import org.junit.After
 import org.junit.Assert.assertEquals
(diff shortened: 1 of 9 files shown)


0xh3rman:
will create a new one

PR #771, opened 2026-07-23
Generate image URLs in core

gemdev111:
Asset, validator, NFT, and list image URLs were built by hand in three places: core, iOS, and Android. URL building now lives in core behind a gemstone `ImageFormatter`, same shape as #753. Generated URLs are unchanged and covered by Rust tests.

## Scope: core + Android only

iOS keeps its pure Swift `AssetImageFormatter` in Primitives untouched — the widget renders token icons through it, and moving it behind gemstone would link the whole Rust library into the appex (the widget binary carries 0 uniffi symbols, verified with nm). The URL format is pinned on both sides by tests: `AssetImageFormatterTests` in Primitives and the core `image_formatter` tests share the same expectations. Android drops its hand-rolled `IconUrlGeneration` and delegates to core.


DRadmir (changes requested):
Does the price widget actually pull in the whole gemstone lib with this, or does the linker strip most of it out?

PR #774, opened 2026-07-24
Move service status latency probe to core

### LINKED ISSUE #766: Move service status latency probe to core
The Service Status screen measures endpoint latency with a hand-rolled HTTP request on both platforms — URLSession on iOS, HttpURLConnection on Android — instead of going through core like the rest of our network calls.

The list of probed endpoints is also declared twice, so the two apps can drift apart on what they actually check.

Move both the latency measurement and the endpoint list into core so each app just renders the result.

DRadmir:
The Service Status screen measured latency with its own HTTP code on each platform, and the list of probed endpoints was written twice, so the apps could drift apart.

Both now live in core, and each app just renders the result.

Also fixes the Android screen showing stale values when you leave it and come back.

Closes #766


0xh3rman (approved):
LGTM

PR #787, opened 2026-07-27
Show dash for empty memo on Android in Confrim

### LINKED ISSUE #785: Show "-" for empty memo on Solana confirmation (Android)
On the send confirmation screen for Solana, iOS always shows the memo row and displays "-" when no memo is set. On Android the memo row is hidden completely in that case. Update Android to show the memo row with "-" when no memo is set, matching iOS.

gemdev111:
closes #785 

img:
image


gemcoder21 (commented):
can you add unit test? make sure we have same on iOS and android unit test

PR #795, opened 2026-07-28
Show WalletConnect approvals with correct type and token

### LINKED ISSUE #763: iOS shows WalletConnect approvals as "Sent" in Activity instead of a smart contract interaction
On iOS, WalletConnect high-amount and unlimited token approvals show up in the Activity list as "Sent 0 ETH". They should be displayed as a smart contract / approval interaction, the way Android already shows them.

Image

DRadmir:
WalletConnect token approvals now show as "Approve" with the approved token in the Activity list, instead of "Sent 0 ETH".

Closes #763

Screenshot_1785229100
Screenshot_1785229104
Simulator Screenshot - wt2 - 2026-07-28 at 16 58 34
Simulator Screenshot - wt2 - 2026-07-28 at 16 58 38

### DIFF
--- a/core/crates/gem_evm/src/transaction.rs
+++ b/core/crates/gem_evm/src/transaction.rs
@@ -1,0 +1,42 @@
+use alloy_primitives::hex;
+use alloy_sol_types::SolCall;
+use primitives::TransactionType;
+
+use crate::contracts::erc20::IERC20;
+
+pub fn decode_transaction_type(input: Option<&str>) -> TransactionType {
+    let calldata = input
+        .map(|value| value.strip_prefix("0x").unwrap_or(value))
+        .and_then(|value| hex::decode(value).ok())
+        .unwrap_or_default();
+    match calldata.as_slice() {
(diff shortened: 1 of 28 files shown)


core/crates/gem_evm/src/transaction.rs
```diff
@@ -0,0 +1,42 @@
+use alloy_primitives::hex;
+use alloy_sol_types::SolCall;
+use primitives::TransactionType;
+
+use crate::contracts::erc20::IERC20;
+
+pub fn decode_transaction_type(input: Option<&str>) -> TransactionType {
```
gemcoder21, line 7: we also need to have `TransactionSmartContractMetadata` for the transation in the future. 

Can you create this as generic protocol so any chain could return proper data?

gemcoder21: or implement this under `WalletConnectRequestHandler` for each chain

gemcoder21 (approved):

PR #796, opened 2026-07-28
Open network assets from the asset screen

DRadmir:
Tapping the Network row on a coin now opens the list of your assets on that network. Available on iOS and Android.

Simulator Screenshot - wt3 - 2026-07-28 at 16 49 02
Simulator Screenshot - wt3 - 2026-07-30 at 15 57 37
Simulator Screenshot - wt3 - 2026-07-30 at 15 57 43


gemdev111 (approved):

gemcoder21 (approved):
that's a good feature to have

consider:
- maybe split between active / hidden assets here?
- do request to fetch all assets for this chain? like a forced method to fetch balances for active / hidden. because hidden assets are not getting updated.

PR #802, opened 2026-07-29
Document how device and subscriptions work across the apps

gemdev111:
Follow-up to #789: there was no reference for how a device registers, when subscriptions are
reconciled, and what has to hold on both platforms, so each fix in this area started by
re-reading two codebases.

`docs/DEVICE_SUBSCRIPTIONS.md` covers the device record and endpoints, the sync flow, the
ordering rule that a wallet is subscribed before any wallet-scoped request, failure and
concurrency behaviour, an iOS/Android mapping with the differences that are deliberate today,
and the invariants a change on either platform has to keep.

It follows `docs/SWAPPER.md` and links to `core/docs` for request signing and streaming
rather than repeating them.

Also adds a Design Docs section to `AGENTS.md`: nothing linked to `docs/`, so neither
`SWAPPER.md` nor this file was discoverable. Happy to drop that part if you would rather keep
the root guide short.

### DIFF
--- a/docs/DEVICE_SUBSCRIPTIONS.md
+++ b/docs/DEVICE_SUBSCRIPTIONS.md
@@ -71,10 +71,4 @@ Keep this document current in the same change when the sync triggers, the reconc
 
 - [iOS device sync](../ios/Packages/FeatureServices/DeviceService/DeviceService.swift)
 - [iOS subscriptions](../ios/Packages/FeatureServices/DeviceService/SubscriptionService.swift)
-- [iOS observer](../ios/Packages/FeatureServices/DeviceService/DeviceObserverService.swift)
-- [iOS mutation sites](../ios/Packages/FeatureServices/WalletService/WalletService.swift)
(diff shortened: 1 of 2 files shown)


docs/DEVICE_SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,80 @@
...
+- Published state is recorded only after a successful sync.
+- Adding a wallet does not remove other wallets' subscriptions; deleting one removes its own.
+- Nothing changed since the last sync means no requests.
+
+Keep this document current in the same change when the sync triggers, the reconcile rules, or the platform mechanisms above change.
+
+## Code map
+
+- [iOS device sync](../ios/Packages/FeatureServices/DeviceService/DeviceService.swift)
+- [iOS subscriptions](../ios/Packages/FeatureServices/DeviceService/SubscriptionService.swift)
```
gemcoder21, line 73: simpler map? only core service, no need mutation sides and so on

gemcoder21 (approved):

PR #822, opened 2026-08-02
Validate stake/transfer amounts from Sui, Aptos, and TON events

DRadmir:
Root-cause fix for a client crash on malformed transaction amounts (follow-up to #818). Sui, Aptos, and TON built the amount for stake/transfer transactions from raw event strings with no digit validation, unlike other numeric fields in the same files. A non-integer or empty value from the upstream event data now makes that transaction get dropped instead of reaching the client broken.

### DIFF
--- a/core/crates/gem_ton/src/provider/transactions_mapper.rs
+++ b/core/crates/gem_ton/src/provider/transactions_mapper.rs
@@ -194,7 +194,7 @@ fn simple_transfer_details(message: &TransactionMessage) -> Option<TransferDetai
             asset_id,
             from: parse_address(&out_message.source)?,
             to,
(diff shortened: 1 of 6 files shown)


core/crates/gem_ton/src/provider/transactions_mapper.rs
```diff
@@ -220,6 +220,14 @@ fn simple_transfer_details(message: &TransactionMessage) -> Option<TransferDetai
     None
 }
 
+fn parse_ton_value(value: &Option<String>) -> String {
```
0xh3rman: we can use BigInt here too? then convert to string instead of this parser / helper

0xh3rman (approved):
LGTM, with one comment

PR #838, opened 2026-08-05
Take perpetual balance from spot state for unified accounts

### LINKED ISSUE #675: Perpetuals balance amount jumps from value to zero
Check the video

https://github.com/user-attachments/assets/31a54f9d-b1fd-4f4e-b10e-b4502f6abf61

maybe the cause of the issue that address has open position and the token is not available in Gem's Perpetuals. 

Image

DRadmir:
Perpetual balance jumped between the real value and zero for wallets with unified account mode enabled. Their trading money lives in the spot balance, while the app read it from the perpetual account, which is empty for such wallets.

Now the balance for those wallets comes from the spot state and the available amount is the one Hyperliquid itself reports. The same money is no longer counted twice in the wallet total, and sending no longer offers the part that backs an open position. Wallets in the standard mode keep working as before.

Closes #675

### DIFF
--- a/core/crates/gem_hypercore/src/provider/websocket_mapper.rs
+++ b/core/crates/gem_hypercore/src/provider/websocket_mapper.rs
@@ -10,14 +10,13 @@ use crate::models::{
 use super::perpetual_mapper::{map_perpetual_balance_from_spot, map_positions, map_tp_sl_from_orders};
 
 pub fn account_subscriptions(address: String, mode: PerpetualAccountMode) -> Vec<HyperliquidSubscription> {
-    let mut subscriptions = vec![
-        HyperliquidSubscription::AccountState { address: address.clone() },
-        HyperliquidSubscription::OpenOrders { address: address.clone() },
-    ];
-    if mode == PerpetualAccountMode::Unified {
-        subscriptions.push(HyperliquidSubscription::SpotState { address });
+    let account_state = HyperliquidSubscription::AccountState { address: address.clone() };
+    let open_orders = HyperliquidSubscription::OpenOrders { address: address.clone() };
+
+    match mode {
+        PerpetualAccountMode::Standard => vec![account_state, open_orders],
+        PerpetualAccountMode::Unified => vec![account_state, open_orders, HyperliquidSubscription::SpotState { address }],
     }
-    subscriptions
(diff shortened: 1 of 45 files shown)


core/gemstone/src/gateway/mod.rs
```diff
@@ -170,6 +170,12 @@ impl GemGateway {
         self.with_provider(chain, |provider| async move { provider.get_positions(address).await }).await
     }
 
+    pub async fn get_perpetual_account_mode(&self, chain: Chain, address: String) -> GemPerpetualAccountMode {
```
gemdev111: Please return `Result` here. Any failed `userAbstraction` request becomes `Standard` - for a unified wallet that means no `spotState` subscription

core/crates/gem_hypercore/src/provider/websocket_mapper.rs
```diff
@@ -1,25 +1,43 @@
...
-    websocket::{ActiveAssetCtxData, HyperliquidSocketMessage, PositionsDiff, RawSocketMessage},
+    websocket::{ActiveAssetCtxData, HyperliquidSocketMessage, HyperliquidSubscription, PositionsDiff, RawSocketMessage},
 };
 
-use super::perpetual_mapper::{map_positions, map_tp_sl_from_orders};
+use super::perpetual_mapper::{map_perpetual_balance_from_spot, map_positions, map_tp_sl_from_orders};
 
-pub fn parse_websocket_data(data: &[u8]) -> Result<HyperliquidSocketMessage, serde_json::Error> {
+pub fn account_subscriptions(address: String, mode: PerpetualAccountMode) -> Vec<HyperliquidSubscription> {
+    let mut subscriptions = vec![
```
gemdev111: can skip let mut

```
  let account_state = HyperliquidSubscription::AccountState { address: address.clone() };
    let open_orders = HyperliquidSubscription::OpenOrders { address: address.clone() };
    match mode {
        PerpetualAccountMode::Standard => vec![account_state, open_orders],
        PerpetualAccountMode::Unified => vec![account_state, open_orders, HyperliquidSubscription::SpotState { address }],
    }
```

gemdev111 (approved):

PR #856, opened 2026-08-08
Use Android Keystore for device auth storage

0xh3rman:
## Summary

- Store new device-auth keys with direct Android Keystore AEAD and a dedicated alias.
- Lazily migrate readable values from the legacy Tink keyset-backed store.
- Reset only the regenerable device-auth values when the direct Keystore key is missing or cannot authenticate the ciphertext.
- Keep the password and Gem preference stores unchanged.

## Root cause

The exact originating Keystore condition is unknown. The crash happens when `AndroidKeysetManager` cannot authenticate its SharedPreferences-wrapped Tink keyset with the Android Keystore master key, which raises `BadPaddingException` or `AEADBadTagException`.

`GetDeviceIdImpl` catches the failed read and generates a new device key pair, but the old implementation then writes through the same unreadable Tink keyset and crashes again. New device-auth writes now use Android Keystore AEAD directly, so regeneration no longer depends on that legacy keyset. Existing readable values migrate on first read.

## Impact

Fixes the two observed device-auth crash paths. Invalid direct-Keystore data is cleared so the device identity can be regenerated; unrelated Android Keystore failures may still surface.

## Validation

- `./gradlew :app:assembleGoogleDebug`
- `./gradlew :app:lintGoogleDebug`
- 8 `TinkEncryptedKeyValueStoreInstrumentedTest` tests on API 35
- `git diff --check`

### DIFF
--- a/android/app/src/main/kotlin/com/gemwallet/android/data/password/EncryptedKeyValueStore.kt
+++ b/android/app/src/main/kotlin/com/gemwallet/android/data/password/EncryptedKeyValueStore.kt
@@ -1,0 +1,141 @@
+package com.gemwallet.android.data.password
+
+import android.content.Context
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.Base64
+
+internal class EncryptedKeyValueStore(
+    context: Context,
+    preferencesFileName: String,
+    private val namespace: String,
+    private val aeadProvider: AeadProvider,
+    private val legacyStore: SecureStringStore,
+    private val resetOnInvalidKey: Boolean,
+) : SecureStringStore {
+
+    private val sharedPreferences = context.applicationContext.getSharedPreferences(
+        preferencesFileName,
+        Context.MODE_PRIVATE,
+    )
+
+    private val namespacePrefix = "${namespace}_"
+
+    override fun contains(key: String): Boolean = sharedPreferences.contains(storageKey(namespace, key))
+
+    override fun getString(key: String): String? {
+        val encryptedValue = sharedPreferences.getString(storageKey(namespace, key), null) ?: return null
+        if (!encryptedValue.startsWith(KEYSTORE_VALUE_PREFIX)) {
+            val legacyValue = readLegacyValue(key) ?: return null
+            runCatching { putString(key, legacyValue) }
+            return legacyValue
+        }
+        val encodedValue = encryptedValue.removePrefix(KEYSTORE_VALUE_PREFIX)
+        return try {
+            decryptKeystoreValue(key, encodedValue)
+        } catch (error: Exception) {
+            recoverStoredValue(key, encodedValue, error)
+        }
+    }
+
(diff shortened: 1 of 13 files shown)


android/app/src/main/kotlin/com/gemwallet/android/data/password/EncryptedKeyValueStore.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    private val namespacePrefix = "${namespace}_"
+
+    override fun contains(key: String): Boolean = sharedPreferences.contains(storageKey(namespace, key))
+
+    override fun getString(key: String): String? {
+        val encryptedValue = sharedPreferences.getString(storageKey(namespace, key), null) ?: return null
+        if (!encryptedValue.startsWith(KEYSTORE_VALUE_PREFIX)) {
+            val legacyValue = readLegacyValue(key) ?: return null
+            runCatching { putString(key, legacyValue) }
```
DRadmir: This swallows the migration failure silently, while the same fallback in `TinkDeviceAuthStore` logs it with `Log.e`. If migration keeps failing on every read, we'd never know. Can we log the error here too?

DRadmir (approved):

android/app/src/main/kotlin/com/gemwallet/android/data/password/EncryptedKeyValueStore.kt
0xh3rman: Added back in b7374130be, mirroring the `TinkDeviceAuthStore` pattern with the namespace for context.

0xh3rman:
Verified the migration and self-heal end-to-end on an emulator upgrading from 2.114.6 (old-format values), which surfaced a bug — fixed in 64e643ae11:

**Bug:** after `resetOnInvalidKey` rotates the device keypair, the device could never re-register. `getOrCreateDevice` short-circuited on the stale local `DeviceRegistered` flag and called `getDevice()`, which throws `GemApiException: Device not found` for the rotated id (`GemApiErrorInterceptor` converts the error body) instead of returning null — so the `registerDevice` fallback was unreachable. The WebSocket looped `Device not found` across app restarts, permanently. Fix: consult backend `isDeviceRegistered()` instead of the local flag.

**Verification (Pixel 10 Pro AVD, API 37):**
- Migration: both `gem_device_keys` values rewritten in place to `android-keystore-v1:` format, no error logs, device id unchanged, second launch idempotent (byte-identical ciphertexts)
- Self-heal: corrupted both stored values → no crash, corrupt entries dropped, keypair regenerated; with the fix the rotated device re-registers (pushed device id updated, subscriptions reconciled, zero WS errors)

Note: `:data:repositories` unit tests currently don't compile on this branch (`AssetsRepositoryTest.kt` out of sync with `AssetsRepository`, 14 errors, pre-existing).

PR #862, opened 2026-08-10
Stop transaction status polling when the wallet is removed

### LINKED ISSUE #850: Transaction status polling crashes the app after wallet deletion
Deleting a wallet while it still has a pending or in-transit transaction leaves the status polling job running with the deleted wallet id. On the next status change the job re-inserts the transaction row and the app crashes with a foreign key constraint error (SQLITE_CONSTRAINT 787), reproduced with a swap and confirmed by a production stack trace. A fix making the polling write path update-only is ready and should land after #843, since both change the same transaction code.

DRadmir:
Deleting a wallet while it still had a pending transaction left the status poller running. On Android the next status update crashed the app, and on iOS the poller kept running forever and processed results for the removed wallet.

Now the poller stops as soon as its transaction is gone, and a status update can no longer crash the app.

Closes #850

### DIFF
--- a/ios/Packages/Primitives/Sources/JobRunner/JobRunner.swift
+++ b/ios/Packages/Primitives/Sources/JobRunner/JobRunner.swift
@@ -52,6 +52,9 @@ extension JobRunner {
                     debugLog("transaction status complete: id=\(job.id), status=complete, error=\(error)")
                 }
                 return
+            case .cancelled:
+                debugLog("transaction status cancelled: id=\(job.id), status=cancelled")
+                return
             case let .retry(error):
                 let sleepUntil = attemptStart.advanced(by: .milliseconds(Int(intervalMs)))
                 if clock.now < sleepUntil {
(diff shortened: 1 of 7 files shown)


ios/Packages/Primitives/Sources/JobRunner/JobRunner.swift
```diff
@@ -52,6 +52,9 @@ extension JobRunner {
                     debugLog("transaction status complete: id=\(job.id), status=complete, error=\(error)")
                 }
                 return
+            case .cancelled:
```
gemcoder21, line 55: maybe just use `.complete` no need for cancel

DRadmir: .complete would run onComplete(), which post processes balances/stake for the wallet from the job's snapshot

gemdev111 (approved):

gemcoder21 (approved):

PR #864, opened 2026-08-10
Own confirm fee state and reload only the fee on speed change

DRadmir:
Fee state on the confirm screen now has a single owner: the confirm view model holds the selection, and the fee scene model is derived from loaded state.

Changing the fee speed now reloads only the fee data — the simulation and balance changes stay on screen.

### DIFF
--- a/ios/Features/Transfer/Sources/ViewModels/ConfirmTransferSceneViewModel.swift
+++ b/ios/Features/Transfer/Sources/ViewModels/ConfirmTransferSceneViewModel.swift
@@ -19,7 +19,7 @@ import WalletConnector
 @Observable
 @MainActor
 public final class ConfirmTransferSceneViewModel {
-    public var feeModel: NetworkFeeSceneViewModel
+    var feeSelection: FeeSelection
     var state: ConfirmTransferState {
         didSet { onStateChange(state: state) }
     }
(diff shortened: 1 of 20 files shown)


ios/Features/Transfer/Sources/ViewModels/ConfirmTransferSceneViewModel.swift
```diff
@@ -286,13 +290,16 @@ extension ConfirmTransferSceneViewModel {
...
-            feeModel.update(feeAmount: data.input.transactionData.fee.fee)
+            if state.feeRates.isEmpty {
+                let data = try await confirmService.load(request: request, selection: feeSelection)
+                state = .loaded(data)
+            } else {
+                let preload = try await confirmService.preload(request: request, selection: feeSelection)
+                state.update(preload)
+            }
         } catch {
+            guard !Task.isCancelled else { return }
```
gemdev111, line 302: don't need  guard !Task.isCancelled else { return }
setError already checks on isCancelled.

gemdev111 (approved):

ios/Features/Transfer/Sources/ViewModels/ConfirmTransferSceneViewModel.swift
DRadmir: setError only filters cancellation errors. This guard covers a different case: the task was cancelled (e.g. fee speed changed), but the request failed with a real error. Without the guard the stale task would write its error over the new task's state.

PR #880, opened 2026-08-12
Integrated "Scan QR code" from wallet

gemdev111:
Adds a QR scanner to the wallet screen, behind developer mode.

Scanned codes are decoded in core - BIP-21, ERC-681, Solana Pay, TON and XRP - so iOS and Android answer the same code the same way instead of each parsing it themselves.

A code carrying a signable amount opens confirm, anything else fills the recipient screen, and a code for another asset or an amount finer than the asset can hold is refused rather than sent as something else.

images:

image
image


core/crates/primitives/src/payment_decoder/erc681.rs
```diff
@@ -1,148 +1,53 @@
...
+const QUERY_ADDRESS: &str = "address";
+const QUERY_AMOUNT: &str = "amount";
+const QUERY_MEMO: &str = "memo";
+const QUERY_VALUE: &str = "value";
+
+pub fn decode(path: &str) -> Result<Payment> {
+    let (path, query) = path.split_once('?').unwrap_or((path, ""));
+    let (target, function) = path.split_once('/').map_or((path, None), |(target, function)| (target, Some(function)));
+    let (target, chain) = match target.split_once('@') {
+        Some((target, network_id)) => (target, chain(network_id).unwrap_or(Chain::Ethereum)),
```
DRadmir: If @chainId is unknown we fallback to Ethereum. Can we throw error instead?

DRadmir (commented):
Should we support WalletConnect for qr feature?

ios/Features/Transfer/Sources/ViewModels/RecipientSceneViewModel.swift
```diff
@@ -251,17 +219,39 @@ extension RecipientSceneViewModel {
...
+            update(
+                from: RecipientData(
+                    recipient: Recipient(name: .none, address: chain.checksumAddress(payment.address), memo: payment.memo),
+                    amount: .none,
+                ),
+            )
+        }
+    }
+
+    private func update(from recipientData: RecipientData) {
```
DRadmir, line 244: Can you check this case: update(from:) sets address first and then amount, but the view has .onChange(of: addressInputModel.text) which calls onChangeAddressText and clears amount

core/crates/primitives/src/payment_decoder/erc681.rs
```diff
@@ -1,148 +1,53 @@
...
+        None => (target, Chain::Ethereum),
+    };
+    let target = target.strip_prefix(PAY_PREFIX).unwrap_or(target);
+    let parameters = query::parameters(query);
+    let memo = query::value(&parameters, QUERY_MEMO);
+
+    match function {
+        Some(TRANSFER_FUNCTION) => Ok(Payment::Request(PaymentRequest {
```
DRadmir, line 35: For token transfers we drop the uint256 amount, so the amount from a merchant QR is silently lost

gemdev111:
> Should we support WalletConnect for qr feature?

Sure, added support in this pr

android/features/assets/presents/src/main/kotlin/com/gemwallet/android/features/assets/views/AssetsTopBar.kt
```diff
@@ -62,6 +64,20 @@ internal fun AssetsTopBar(
                 }
             }
         },
+        navigationIcon = {
+            if (onScan != null) {
+                IconButton(
+                    onClick = onScan,
+                    Modifier.testTag("assetsScanAction")
```
gemcoder21: for test tags, can you create constants for all tags?

core/crates/primitives/src/payment_decoder/bip21.rs
```diff
@@ -0,0 +1,71 @@
...
+        return Ok(Payment::Request(PaymentRequest {
+            address: path.to_string(),
+            amount: None,
+            memo: None,
+            asset_id,
+        }));
+    };
+
+    let parameters = query::parameters(query);
+    if let Some(required) = parameters.keys().find(|key| key.starts_with(REQUIRED_PARAMETER_PREFIX)) {
```
gemcoder21, line 38: why do you check for `REQUIRED_PARAMETER_PREFIX` in bip21?

core/crates/primitives/src/payment_decoder/erc681.rs
```diff
@@ -1,148 +1,59 @@
...
-
-        Ok(TransactionRequest {
-            target_address,
-            prefix,
-            chain_id,
-            function_name,
-            parameters,
-        })
+const PAY_PREFIX: &str = "pay-";
+const TRANSFER_FUNCTION: &str = "transfer";
```
gemcoder21, line 12: why tests were removed for erc681? it should stay here just for this spec

core/crates/primitives/src/payment_decoder/ton_pay.rs
```diff
@@ -1,62 +1,40 @@
...
+
+const QUERY_AMOUNT: &str = "amount";
+const QUERY_TEXT: &str = "text";
+const QUERY_BODY: &str = "bin";
+const QUERY_STATE_INIT: &str = "init";
+
+pub fn decode(path: &str) -> Result<Payment> {
+    let (path, query) = path.split_once('?').unwrap_or((path, ""));
+    let parameters = query::parameters(query);
+    query::reject_unsupported(&parameters, &[QUERY_BODY, QUERY_STATE_INIT])?;
```
gemcoder21: remove `reject_unsupported`, this is not needed for all specs

core/crates/primitives/src/chain.rs
```diff
@@ -68,6 +68,13 @@ pub enum Chain {
     Stable,
 }
 
+pub(crate) fn network_id_value(network_id: &str) -> Option<u64> {
```
gemcoder21: `network_id_value` code smell

core/crates/primitives/src/payment.rs
```diff
@@ -0,0 +1,52 @@
...
+pub enum Payment {
+    Request(PaymentRequest),
+    Link(PaymentLink),
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[typeshare(swift = "Equatable, Hashable, Sendable")]
+#[serde(tag = "type", content = "content", rename_all = "camelCase")]
+pub enum PaymentAmount {
+    ExactValue(String),
```
gemcoder21, line 19: `String` is this BigInt value? can we use big int in this case?

ios/Features/Transfer/Sources/ViewModels/RecipientSceneViewModel.swift
```diff
@@ -33,7 +34,7 @@ public final class RecipientSceneViewModel {
     public var isPresentingScanner: RecipientScene.Field?
     var addressInputModel: AddressInputViewModel
     var memo: String = ""
-    var amount: String = ""
+    private(set) var scanned: RecipientData?
```
gemcoder21: `scanned` need better naming

ios/Gem/Navigation/NavigationHandler.swift
```diff
@@ -43,26 +58,34 @@ final class NavigationHandler: Sendable {
...
+        await handle(action)
+    }
+
     @MainActor
     func handle(_ action: URLAction) async {
         do {
             try await handleURLAction(action)
         } catch {
             debugLog("NavigationHandler URLAction error: \(error)")
+            presentNotSupported()
```
gemcoder21: `presentNotSupported` code smell, should be handler for showing an error instead, `showError(error)`

ios/Gem/Navigation/NavigationHandler.swift
```diff
@@ -93,6 +117,43 @@ extension NavigationHandler {
...
 
+// MARK: - Payment
+
+@MainActor
+extension NavigationHandler {
+    private func handlePayment(_ payment: Payment) throws {
+        guard case let .request(request) = payment, let wallet else {
+            throw AnyError(Localized.Errors.notSupported)
+        }
+        let assets = try assetsService.assetStore.getAssetsData(walletId: wallet.id, filters: [.enabledBalance])
```
gemcoder21: why does it have to be enabled filter?

ios/Packages/PrimitivesComponents/Sources/Types/PaymentAsset.swift
```diff
@@ -0,0 +1,27 @@
+// Copyright (c). Gem Wallet. All rights reserved.
+
+import Foundation
+import GemstonePrimitives
+import Primitives
+
+enum PaymentAsset {
+    case unsupported
+    case single(AssetData)
```
gemcoder21: this should be just `Asset` or `AssetId` or list of Chains to show, which will include any asset

ios/Packages/PrimitivesComponents/Sources/Types/PaymentTransfer.swift
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+}
+
+// MARK: - Private
+
+private extension PaymentTransfer {
+    func isSameAsset(_ payment: PaymentRequest) -> Bool {
+        guard let assetId = payment.assetId else { return true }
+        return assetId == asset.id
```
gemcoder21: do you need guard here?

core/crates/primitives/src/payment_decoder/bip21.rs
gemdev111: BIP-21(https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki ) requires it - if a client doesn't implement a `req-` param it must treat the whole URI as invalid. Only BIP-21 has that rule, so it's only here.

core/crates/primitives/src/payment.rs
gemdev111: `ExactValue` is a decimal in whole units - it needs the asset's decimals to become a BigInt, and only the app has those for tokens (`solana:…?amount=1&spl-token=<mint>` = 1 USDC). So core hands over the decimal and the app scales it. `AtomicValue` is already the smallest unit → `BigUint`.

PR #883, opened 2026-08-13
Disable SEI cosmos chain

0xh3rman:
Disables Sei Cosmos while preserving its Rust enum and database compatibility.

- Marks Sei variants with `#[typeshare(skip)]`, so generated iOS and Android enums no longer expose the discontinued chain while Core can still decode existing values.
- Adds one-time Android 85→86 and iOS database migrations using the raw `"sei"` identifier to transactionally remove chain references.
- Deletes wallets left without accounts through the normal wallet deletion flows, including private keys, preferences, and avatars. Multi-chain wallets remain.
- Keeps Sei EVM assets intact under the explicit `seievm` name.

The migration helpers accept a set of raw chain identifiers, providing a generic method to disable future chains without adding client enum cases or permanent runtime filtering.

Validated with the Android Store migration test and assembly, plus the iOS Store test suite.


DRadmir (commented):
@0xh3rman

Sei is only filtered in two places (allChains on iOS and available() on Android), but many places still use Chain.allCases / Chain.entries directly for example token search, transactions filter, contact chain select. So Sei can still appear there.

I suggest a different approach: add a "discontinued" flag to the chain config in core, and write a generic cleanup on each platform that runs on app start: for every discontinued chain it deletes accounts/addresses/nodes and disables assets. This cleanup is data-only, the schema does not change, so we don't need DB migrations at all.

We write this logic once, and next time we disable a chain it is only one flag in core no new migrations

Also, deleting accounts can leave a wallet with zero accounts

gemcoder21 (changes requested):
it should be generic solution, for any chain

android/data/services/store/src/androidTest/kotlin/com/gemwallet/android/service/store/Migration_85_86Test.kt
```diff
@@ -0,0 +1,126 @@
...
+                '0', 0.0,
+                '0', 0.0,
+                '0', 0.0,
+                '0', 0.0,
+                '0', 0.0,
+                '0', 0.0,
+                '0', 0.0,
+                1.0, 1, 1, 1, 0, 0
+            )
+            """.trimIndent()
```
gemcoder21: `.trimIndent()` why?

android/data/services/store/src/androidTest/kotlin/com/gemwallet/android/service/store/Migration_85_86Test.kt
```diff
@@ -0,0 +1,126 @@
...
+import com.gemwallet.android.data.service.store.database.GemDatabase
+import com.gemwallet.android.data.service.store.database.di.Migration_85_86
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class Migration_85_86Test {
```
gemcoder21, line 18: just unit test to test removal of all references, make sure they are in the right order

0xh3rman:
@DRadmir we still need to have a place to filter `discontinued` flag and most chains are false, now we have a disabled chain config and replaces all the `Chain.allCases` with `Chain.enabled`

DRadmir (commented):
We delete wallets with raw SQL, but this skips the normal wallet delete flow (WalletService.delete on iOS, DeleteWalletImpl on Android). So for Sei-only wallets we leave data: private keys, wallet preferences, and the avatar too. Maybe we should delete these wallets through the normal wallet delete path?

Small thing: Chain::all() in Rust is not filtered, so core/backend still see Sei as enabled. Is that intended?

0xh3rman:
@DRadmir good suggestion, let me apply

gemcoder21 (changes requested):
This should handle removal of chain case on the clients, sei should be marked as skip generation for typeshare. Once user launches the app it removed any references of that chain once, and never checks inside the app

gemcoder21:
once that change is made, there is many others that does not need renaming or adjustments, less code and much more reliable and long term solution to remove any unused stuff.

android/data/services/store/src/main/kotlin/com/gemwallet/android/data/service/store/database/di/Migration_85_86.kt
```diff
@@ -0,0 +1,76 @@
...
+}
+
+internal data class MigrationStatement(
+    val sql: String,
+    val arguments: List<Any?>,
+) {
+    override fun toString(): String = "$sql | ${arguments.joinToString()}"
+}
+
+internal fun chainRemovalStatements(chain: String): List<MigrationStatement> {
```
gemcoder21: `chainRemovalStatements` move this to reusable function somewhere? so could be used in the future

ios/Packages/Store/Sources/Migrations.swift
```diff
@@ -23,6 +25,62 @@ struct Migrations {
         try? db.execute(sql: "DELETE FROM \(AssetRecord.databaseTableName) WHERE chain = ?", arguments: [chain])
     }
 
+    static func removeChains(_ db: Database, chains: Set<String>) throws {
```
gemcoder21: there is `clearChainData` - can you consolidate that? should we check if table exist first?

core/crates/gem_evm/src/uniswap/deployment/v4.rs
```diff
@@ -175,7 +175,7 @@ mod tests {
 
     #[test]
     fn test_deployment_addresses_are_checksummed() {
-        for chain in Chain::all() {
+        for chain in Chain::all_with_disabled() {
```
gemcoder21: can we remove `all_with_disabled`? just exclude sei in `all()`

0xh3rman:
close for now, too many conflicts and better re-impl with a clean state

PR #892, opened 2026-08-14
Derive current wallet instead of copying it into every screen

### LINKED ISSUE #833: Wallet state is copied into every screen instead of derived
Each screen keeps its own `var wallet` and updates it by hand from a separate `.onChange` observer. The rule for when a wallet change should re-scope the screen's queries versus just redraw it is rewritten in four view models, and all four versions differ:

```swift
// WalletSceneViewModel — diffs id, then value
if wallet.id != newWallet.id { refresh(for: newWallet) }
else if wallet != newWallet { wallet = newWallet }

// TransactionsViewModel — always full refresh
if let newWallet, wallet != newWallet { refresh(for: newWallet) }

// CollectionsViewable — value only, no id check
if let newWallet, wallet != newWallet {
    wallet = newWallet
    query.request = NFTRequest(walletId: newWallet.id, filter: .all)
}

// MainTabViewModel — no diff at all
wallet = newWallet
transactionsQuery.request.walletId = newWallet.id
```

The source they all read is only half observable, so renaming a wallet invalidates nothing:

```swift
var currentWallet: Wallet? {
    guard let currentWalletId else { return nil }           // observable
    return try? walletStore.getWallet(id: currentWalletId)  // plain read, not observed
}
```

A rename reaches the wallet tab only because closing the rename sheet happens to trigger a redraw that re-reads it.

Make the current wallet one derived value that screens read, so a rename propagates by itself and switching a wallet rebuilds the screen instead of being patched into it.

Entry points: `onChangeWallet`, `refresh(for:)`, `CollectionsViewable`, `WalletSessionService.currentWallet`.

DRadmir:
Current wallet is derived from a single source instead of being copied into every screen. Switching wallets rebuilds per-wallet screens automatically, and opening a push notification for another wallet now lands on the right screen reliably.

Closes #833


ios/Features/ManageWallets/Sources/ViewModels/WalletIDetailViewModel.swift
```diff
@@ -23,7 +23,7 @@ public final class WalletDetailViewModel {
     var isPresentingExportWallet: ExportWalletType?
 
     public let walletQuery: ObservableQuery<WalletRequest>
-    public var dbWallet: Wallet? {
+    public var dbWallet: Wallet {
```
gemcoder21:
```suggestion
    public var wallet: Wallet {
```

can we do this then?

gemcoder21 (approved):

PR #908, opened 2026-08-18
Refactor gem_evm and move bsc/monad/everstake code into their own crate

0xh3rman:


gemcoder21:
Let's re-open once Tempo is done and come back to this.

PR #931, opened 2026-08-20
Remove tag search from asset selection

DRadmir:
Removes search by tags. The row of tag chips is gone from the wallet search and from every asset picker — send, swap, buy, receive and manage assets.

Asset lists are untouched and keep working as before.

Simulator Screenshot - wt2 - 2026-08-20 at 15 33 34

Screenshot_1787211774


gemcoder21 (approved):

gemcoder21:
@DRadmir we might need those tags views for the explore / discover feature, so maybe some views ok to keep

PR #991, opened 2026-08-27
Add a Scan and Receive switcher to the wallet scanner

### LINKED ISSUE #913: Extend Scan to Support Receive

gemdev111:
The wallet scanner becomes a Scan | Receive switcher: scan a code or show your own QR from the same sheet. The Receive segment is the receive asset list - search, network filter, and recents - and picking an asset presents its QR on top, keyed per asset.

- Segmented Scan | Receive toggle on both platforms; ✕ closes the scanner, the QR sheet closes back to the list
- The scanner flow is locked to portrait on both platforms
- Receive filters and search reset when the scanner closes, so one session's filters don't leak into the next

Fixes along the way:

- Rotating the phone closed the Android scanner
- The iOS scanner started the camera during a SwiftUI view update
- Recents sheets flashed the empty state while closing (also fixes send and buy recents)

| | iOS | Android |
|---|---|---|
| Receive | ios-receive | android-receive |
| Asset QR | ios-qr | android-qr |

closes #913


ios/Gem/Types/ScanReceiveMode.swift
```diff
@@ -0,0 +1,8 @@
+// Copyright (c). Gem Wallet. All rights reserved.
+
+import Foundation
+
+enum ScanReceiveMode: CaseIterable, Sendable {
```
gemcoder21: move to primitives

ios/Gem/Types/ScanReceiveSheetType.swift
```diff
@@ -0,0 +1,15 @@
+// Copyright (c). Gem Wallet. All rights reserved.
+
+import Primitives
+
+enum ScanReceiveSheetType: Identifiable {
```
gemcoder21: `ScanReceiveSheetType` not needed?

gemcoder21 (approved):

Issue #855, opened 2026-08-08
Expand chain coverage for Relay swap


0xh3rman:
- [x] TRON
- [x] Solana
- [x] TON

gemcoder21:
@0xh3rman what other chains we should support for cross chain swaps via relay?

0xh3rman:
@gemcoder21 Solana might be a good option

gemcoder21:
@0xh3rman Tron, Solana, Ton is good to cover.

PR #994, opened 2026-08-28
Add TRON swap support to Relay

0xh3rman:
Part of https://github.com/gemwalletcom/wallet/issues/855, tested txs:
- TRX -> ETH (Arbitrum) https://relay.link/transaction/0x1787899730ecb52dcd86bed296d4a9e12b17e2d3e294fa3334e723db7d3171c1
- ETH -> TRX https://relay.link/transaction/0x17878998788ad729a333cb795b23db456583e10f5b05d39393a00ca9490df2d0
- xDAI -> Tron USDT https://relay.link/transaction/0x178790004872c7521dd42d0ceaef1fd32af1d077deb176f0b9199cba3a687f0d
- Tron USDT -> ETH https://relay.link/transaction/0x178790014863a7da20d2ff4d3e753607ca03c38ca116a4bc3fc4997724f66086

simulator_screenshot_995CCCEC-F236-4FF6-B221-6401D3B88E5C Simulator Screenshot - iPhone 17 Pro - 2026-08-28 at 15 51 44
simulator_screenshot_76C95D47-BE3C-48DC-94C5-CDCCFBBC83BF simulator_screenshot_0035C35A-400C-4AB1-B0A0-6DE411FC9E05


core/crates/primitives/src/contract_constants.rs
```diff
@@ -1,4 +1,5 @@
 pub const EVM_ZERO_ADDRESS: &str = "0x0000000000000000000000000000000000000000";
+pub const TRON_ZERO_ADDRESS: &str = "T9yD14Nj9j7xAB4dbGeiX9h8unkKHxuWwb";
```
gemcoder21: why this is called `TRON_ZERO_ADDRESS` ?

core/crates/swapper/src/relay/client.rs
```diff
@@ -35,3 +35,27 @@ where
         self.client.get("/chains").await.map_err(SwapperError::from)
     }
 }
+
+fn map_quote_error(error: ClientError) -> SwapperError {
```
gemcoder21: we need generic solution, to avoid mapping in every provider like this, need to have a way to map success model and error model, simple interface and reuse for all

gemcoder21 (approved):

core/crates/primitives/src/contract_constants.rs
0xh3rman: it's 0x41 prefix on twenty zero bytes, like the EVM zero address in base58 check, tronscan labels it as black hole address, let me change this

core/crates/swapper/src/swapper.rs
```diff
@@ -187,18 +199,55 @@ impl GemSwapper {
...
+                Err(error) => errors.push(error),
             }
         }
 
         Self::sort_quotes_by_output_amount(&mut quotes);
-        Ok(SwapQuotes { quotes, errors })
+        Ok((quotes, errors))
+    }
+
+    fn aggregate_quote_error(errors: Vec<(String, SwapperError)>) -> SwapperError {
```
gemcoder21: code smell? re-structure the code, something is wrong and we don't need to round up

core/crates/gem_client/src/types.rs
```diff
@@ -16,6 +16,15 @@ pub enum ClientError {
     Serialization(String),
 }
 
+impl ClientError {
+    pub fn decode_body<T: DeserializeOwned>(&self) -> Option<T> {
```
gemcoder21: can we make it easier to allow passing `T: DeserializeOwned` and then `E: DeserializeOwned` as error type? so it would fallback and tries to decode and error to cover this case?

core/crates/swapper/src/swapper.rs
0xh3rman: app shows "Use Minimum Amount  0.001885 ETH", the real amount is 0.0018854..., if user types `0.001885`, it still below the mininum amount,  need to round up in both android and ios side?

0xh3rman: removed

core/crates/gem_client/src/types.rs
0xh3rman: did some refactoring, now we have ext method to get_or_error if you need to pass an error type, also ClientError::Http body becomes generic

PR #996, opened 2026-08-28
Fix wallet creation and import on Android

DRadmir:
Creating or importing a wallet on Android stopped working for anyone who already had a wallet — the app failed with a "Password not found" error. The rule that decides when a new keystore password may be created was written for iOS, where all wallets share one password, and it does not fit Android, where every wallet has its own.

Each app now decides that for itself: Android creates a password for the new wallet as before, and iOS still refuses to replace the password that protects existing wallets.


gemcoder21 (approved):
Right fix, and in the right place.

The regression came from df70459470, which moved iOS's shared-password rule (can_create_password: no keystore wallet exists yet) into Core. Android gives every wallet its own password, so the second wallet arrived with create_if_missing = false and getPassword rethrew "Password not found".

Letting each adapter decide is the correct layering: the differing fact is the password model, which lives in the GemKeystorePassword implementation, not in the wallet list. Checking the keystore itself via has_stored_wallets is also stronger than the old wallet-list rule, and the password.isNotEmpty short-circuit keeps normal multi-wallet import on iOS working.

Two follow-ups, neither blocking: create_if_missing is now a literal true at its only call site and should leave the trait; and the new iOS test would be worth extending with a second import while the password is present, so a future reordering of that guard cannot pass silently.

gemcoder21:
Correction to my review: ignore the first follow-up. I said create_if_missing was a literal true at its only call site, but there are two — import_wallet passes true and setup_chains passes false, so the flag still carries a real distinction (importing may create a password; adding chains to an existing wallet must never). It stays. The second follow-up, extending the iOS test with an import while the password is present, still holds and I am adding it on main.

PR #1004, opened 2026-08-31
Fix Polkadot available balance

DRadmir:
Polkadot showed locked coins as spendable and hid spendable coins behind a large reserve: an account with everything locked by vesting offered its full balance to Max, while an account with 250 transferable DOT showed a zero balance and nothing to send.

The app now uses the transferable amount reported by the node, and locked coins stay visible in the total instead of disappearing.


core/crates/gem_polkadot/src/models/account.rs
```diff
@@ -1,13 +1,49 @@
...
+    fn test_deserialize_balance_info() {
+        let modern: PolkadotAccountBalance = serde_json::from_str(r#"{"nonce":"48","free":"31415926535","reserved":"0","frozen":"31415926535","transferable":"0"}"#).unwrap();
+
+        assert_eq!(modern.free, BigInt::from(31415926535_u64));
+        assert_eq!(modern.reserved, BigInt::from(0));
+        assert_eq!(modern.frozen, Some(BigInt::from(31415926535_u64)));
+        assert_eq!(modern.transferable, Some(BigInt::from(0)));
+        assert_eq!(modern.nonce, 48);
+
+        let old_runtime: PolkadotAccountBalance =
```
0xh3rman: do we really need to support old runtime? if not we can make the `PolkadotAccountBalance` code simpler by dropping Option<>

core/crates/gem_polkadot/src/models/account.rs
```diff
@@ -1,13 +1,49 @@
...
     pub reserved: BigInt,
+    #[serde(default, serialize_with = "serialize_option_bigint", deserialize_with = "deserialize_bigint_or_none")]
+    pub frozen: Option<BigInt>,
+    #[serde(default, serialize_with = "serialize_option_bigint", deserialize_with = "deserialize_bigint_or_none")]
+    pub transferable: Option<BigInt>,
     #[serde(deserialize_with = "deserialize_u64_from_str")]
     pub nonce: u64,
 }
+
+fn deserialize_bigint_or_none<'de, D>(deserializer: D) -> Result<Option<BigInt>, D::Error>
```
0xh3rman: this should be moved to shared place if we really need it

core/crates/gem_polkadot/src/models/account.rs
DRadmir: It is not about old runtime. Users can select a custom node, and older Sidecar versions return text like "frozen does not exist for this runtime" instead of a number. Without Option the whole response fails to parse and the balance stops updating. With Option we just use the fallback formula. So I would keep it.

0xh3rman: I remember we have some node check? if one node is old sidecar, we should probably reject adding it somehow?

core/crates/gem_polkadot/src/provider/balances_mapper.rs
```diff
@@ -1,32 +1,68 @@
 use crate::models::account::PolkadotAccountBalance;
-use num_bigint::BigUint;
+use num_bigint::{BigInt, BigUint};
 use primitives::{AssetBalance, Balance, Chain};
 
 pub fn map_coin_balance(balance: PolkadotAccountBalance) -> AssetBalance {
-    let available = std::cmp::max(&balance.free - &balance.reserved, num_bigint::BigInt::from(0));
+    let zero = BigInt::from(0);
+    let transferable = balance
```
0xh3rman: we can move this logic to `PolkadotAccountBalance`?

core/crates/gem_polkadot/src/provider/balances_mapper.rs
```diff
@@ -1,32 +1,68 @@
...
+        },
     )
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
 
+    fn account_balance(free: u64, reserved: u64, frozen: Option<u64>, transferable: Option<u64>) -> PolkadotAccountBalance {
+        PolkadotAccountBalance {
```
0xh3rman: this can be moved to testkit and mock_ prefix

PR #1007, opened 2026-08-31
Count the owner reserve in the XRP available balance

DRadmir:
XRP accounts with trustlines or other ledger objects showed a higher available balance than the network actually allows to send, so Max transfers failed with an insufficient reserve error. The reserve now includes the per-object part, matching what the ledger enforces.

### DIFF
--- a/core/crates/gem_xrp/src/provider/balances_mapper.rs
+++ b/core/crates/gem_xrp/src/provider/balances_mapper.rs
@@ -74,33 +74,22 @@ mod tests {
     use crate::models::rpc::AccountInfo;
     use primitives::{AssetId, Chain};
 
-    fn account_info(balance: u64, owner_count: u32) -> AccountInfo {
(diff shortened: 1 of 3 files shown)


core/crates/gem_xrp/src/provider/balances_mapper.rs
```diff
@@ -71,46 +74,35 @@ mod tests {
     use crate::models::rpc::AccountInfo;
     use primitives::{AssetId, Chain};
 
-    #[test]
-    fn test_map_native_balance() {
-        let account = AccountInfo {
-            balance: 10000000, // 10 XRP
+    fn account_info(balance: u64, owner_count: u32) -> AccountInfo {
```
0xh3rman: add mock_ prefix and move to testkit

0xh3rman (approved):

PR #1010, opened 2026-08-31
Create the device key only once on first launch

gemdev111:
The app creates one key for the device on first launch and signs every API request with it. This is not a wallet key — no funds involved.

Two parts of the app can ask for that key at the same moment. Both saw no key, both created one, and both saved it — the keychain keeps the last save. So one caller ended up signing with a key that was no longer stored, and the device id changed on the next launch.

Now the create path runs once: the first caller creates the key, everyone else gets that one.

**This is a regression from `a4e78342fa`.** The Kotlin it replaced (`GetDeviceIdImpl`) did the second check inside the lock; the Rust port dropped it. Android only — on iOS `ServicesFactory` asks once at startup, before anything can race it.

Test: eight callers race the first read. Fails on the old code, passes now.

### DIFF
--- a/core/gemstone/src/services/device/keys.rs
+++ b/core/gemstone/src/services/device/keys.rs
@@ -25,14 +25,15 @@ impl GemDeviceKeyService {
     }
 
     pub fn key_pair(&self) -> Result<GemDeviceKeyPair, GemServiceError> {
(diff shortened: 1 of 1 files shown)


gemcoder21:
Applied directly to main in [3f84fd53c4](https://github.com/gemwalletcom/wallet/commit/3f84fd53c4).

PR #1139, opened 2026-09-04
Give each fiat purchase attempt its own identity

### LINKED ISSUE #1039: Fiat transaction history collapses retried purchases into the oldest attempt
The backend stores one fiat transaction row per provider attempt and several attempts can share one quote id, but both apps use the quote id as the row identity. Because the server returns newest first and the apps write in that order, the oldest attempt overwrites the completed one. A purchase that was declined once and then completed shows as Failed and the completed purchase never appears in the list.

DRadmir:
A purchase that was declined once and then completed showed up in the history as Failed, and the completed purchase never appeared at all. Every attempt of the same purchase was treated as one and the same record, so the older attempt overwrote the newer one.

Each attempt now has its own identity, and the history keeps exactly the set of purchases the server reports for the wallet.

Closes #1039


gemcoder21 (commented):
I think we should return `id` on the backend side

DRadmir:
@gemcoder21 already in this PR the backend returns the row id now instead of the quote id

android/data/services/store/src/main/kotlin/com/gemwallet/android/data/service/store/database/FiatTransactionsDao.kt
```diff
@@ -13,7 +13,16 @@ import kotlinx.coroutines.flow.Flow
 interface FiatTransactionsDao {
 
     @Insert(entity = DbFiatTransaction::class, onConflict = OnConflictStrategy.REPLACE)
-    fun insert(transactions: List<DbFiatTransaction>)
+    suspend fun insert(transactions: List<DbFiatTransaction>)
+
+    @Query("DELETE FROM fiat_transactions WHERE walletId = :walletId AND id NOT IN (:ids)")
+    suspend fun deleteExcept(walletId: String, ids: List<String>)
```
gemcoder21, line 19: `deleteExcept` not used?

DRadmir: It is, setFiatTransactions right below calls it together with insert in one transaction

PR #1142, opened 2026-09-04
Fix Android crash when unstaking Sui delegation

### LINKED ISSUE #1129: Unstaking on Sui crashes android app
Opening Unstake for a Sui delegation crashes the app. Not reproduced on a device yet, so the cause is not confirmed: the amount screen throws when the delegation behind the opened screen cannot be found, and delegation lookups in this area are already known to drop the wallet id. Reproduce it, take the stack trace and fix the layer that lets the screen open without the delegation it needs.

0xh3rman:
## Summary
- serialize direct Unstake, Withdraw, and Rewards variants through the base `StakeType` serializer
- preserve the tagged-enum discriminator expected by the Gemstone JSON bridge
- add a regression test covering discriminator presence and round-trip decoding

## Root cause
The reproduced 2.114.12 device crash was an `uniffi.gemstone.InternalException` from `GemStakeService.stakeTransferData`: Core could not lift the `stake_type` JSON because the payload was missing the `type` field.

Android called the generic `toJson()` extension directly on `StakeType.Unstake`. Kotlin therefore inferred the concrete subtype serializer, which encoded only the variant content. Core deserializes this payload as the Rust tagged `StakeType` enum and requires a discriminator such as `"type":"Unstake"`, so it rejected the payload before transaction construction.

The same inference bug existed in the direct Withdraw and Rewards paths. The amount-screen path was already safe because it first widened the value to `StakeType`. The fix makes that base type explicit at each direct JSON boundary with `toJson<StakeType>()`.

This is independent of ProGuard/R8 and of the original missing-delegation hypothesis in the issue.

## Verification
- `./gradlew :gemcore:testDebugUnitTest --tests com.gemwallet.android.domains.stake.StakeTypeSerializationTest :features:earn:delegation:viewmodels:testDebugUnitTest`
- `SKIP_SIGN=true BUILD_MODE=release GEMSTONE_ANDROID_ABIS=arm64-v8a ./gradlew -Pchannel=google :app:assembleGoogleRelease --no-configuration-cache --no-daemon`

Fixes #1129

### DIFF
--- a/android/features/earn/delegation/viewmodels/src/main/kotlin/com/gemwallet/android/features/earn/delegation/viewmodels/DelegationViewModel.kt
+++ b/android/features/earn/delegation/viewmodels/src/main/kotlin/com/gemwallet/android/features/earn/delegation/viewmodels/DelegationViewModel.kt
@@ -127,7 +127,14 @@ class DelegationViewModel @Inject constructor(
             buildUndelegate()?.let { amountCall(it) }
             return
         }
-        confirmCall(stakeService.stakeTransferData(assetInfo.asset.toGem(), StakeType.Unstake(delegation).toJson(), BigInteger(delegation.base.balance), false))
+        confirmCall(
+            stakeService.stakeTransferData(
+                asset = assetInfo.asset.toGem(),
+                stakeType = StakeType.Unstake(delegation).toJson<StakeType>(),
+                value = BigInteger(delegation.base.balance),
+                useMaxAmount = false,
+            )
+        )
     }
 
     fun onRedelegate(call: AmountTransactionAction) {
@@ -137,7 +144,14 @@ class DelegationViewModel @Inject constructor(
     fun onWithdraw(call: ConfirmTransactionAction) {
         val assetInfo = assetInfo.value ?: return
         val delegation = delegation.value ?: return
-        call(stakeService.stakeTransferData(assetInfo.asset.toGem(), StakeType.Withdraw(delegation).toJson(), BigInteger(delegation.base.balance), false))
+        call(
+            stakeService.stakeTransferData(
+                asset = assetInfo.asset.toGem(),
+                stakeType = StakeType.Withdraw(delegation).toJson<StakeType>(),
+                value = BigInteger(delegation.base.balance),
+                useMaxAmount = false,
+            )
+        )
     }
 
     fun onClaimRewards(call: ConfirmTransactionAction) {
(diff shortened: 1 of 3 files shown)


gemcoder21:
Your root-cause analysis is exactly right, and the reproduction is spot on: `StakeType.Unstake(delegation).toJson()` infers `T` as the concrete subtype, kotlinx.serialization uses that subtype's serializer, and the payload loses the `"type"` discriminator the Rust tagged enum requires. It compiles, it produces valid JSON, and Core rejects it at runtime.

I have landed a fix on `main` in c7abce380c that closes the same hole a level down, so this PR is superseded.

**Why not annotate the call sites**

`toJson<StakeType>()` fixes the four sites we know about. What it does not do is stop the next one — and the reason this bug existed at all is instructive: `PerpetualMappers.kt` already had `fun PerpetualType.toGem(): GemPerpetualType = toJson()`, whose receiver is the base type, so inference cannot pick a subtype there. `StakeType` simply never got that mapper. The difference between a safe tagged type and a crashing one was whether someone had remembered to hand-write a base-typed accessor.

**What landed instead**

`just generate-models` now emits one overload per tagged bridge type, into `android/gemcore/.../serializer/TaggedJsonBridge.kt`:

```kotlin
fun StakeType.toJson(): String = jsonEncoder.encodeToString<StakeType>(this)
```

It reads the `json_bridge!` list in `core/gemstone/src/models/json_bridge.rs` and picks out the enums with data-carrying variants, so it currently covers all 11 — `EarnType`, `Payment`, `PaymentAmount`, `PaymentLink`, `PerpetualModifyPositionType`, `PerpetualType`, `SolanaNftStandard`, `StakeType`, `StreamEvent`, `StreamMessage`, `TronStakeData` — and a type added to that list in future is covered without anyone remembering.

Because the overload is more specific than the generic `<T> T.toJson()`, Kotlin resolves to it and **every existing call site is correct as written** — the four in this PR included, with no diff at those sites.

**Regression test**

`TaggedJsonBridgeTest` in `gemcore` asserts the discriminator survives for variants built inline, that they round-trip back to the right variant, and that widening by hand and letting the overload widen produce identical JSON. I checked it fails without the generated file, so it guards the actual mechanism rather than restating it.

Verified: Core 472 tests and clippy clean; Android `just test`, `assembleGoogleDebug` and `assembleGoogleDebugAndroidTest`; full iOS suite green.

Thanks for chasing this one down to the serializer — the diagnosis is what made the general fix possible.

gemcoder21:
Superseded by c7abce380c on main — the generated base-typed overloads make these four call sites correct as written, and cover the other ten tagged bridge types at the same time.

Issue #1036, opened 2026-09-02
Show Permit approving token


gemcoder21:
<img width="320"  alt="Image" src="https://github.com/user-attachments/assets/ddb12783-4d09-4e6f-aa60-c6ec25d88bd5" />

PR #1169, opened 2026-09-07
Name the known swap contracts by provider and role

0xh3rman:
Part of #1036, new Permit2 approvals look consistent (EIP712 and normal ERC20 Approve), our contract labeling shows both Uniswap (address), one of them is Permit2, another one is Router, this pr adds our known swapper contracts to scan address DB

Changes:
1. Store scan address names as "<provider> Permit2" and "<provider> Router" so the spender row tells Uniswap's Permit2 apart from its Universal Router instead of showing "Uniswap" for both.
2. Upsert the names on the (chain, address) key so a rename reaches rows already in the database on the next setup run; the fraud and verification flags stay untouched.
3. Tag the static THORChain and Maya routers, Chainflip vaults, NEAR Intents treasuries, Mayan contracts and the Squid multicall with their chain at the source, so the provider and the setup step read one table.
4. Add those addresses and the Across multicall handlers under the provider name, so history and confirmation rows show "THORChain" or "Chainflip" instead of a raw address. EVM contracts deployed at one address on every chain sit under Ethereum; the address name lookup already falls back across chains.

### DIFF
--- a/core/crates/storage/src/repositories/scan_addresses_repository.rs
+++ b/core/crates/storage/src/repositories/scan_addresses_repository.rs
@@ -10,6 +10,7 @@ pub trait ScanAddressesRepository {
     fn get_scan_addresses(&mut self, queries: &[(Chain, &str)]) -> Result<Vec<ScanAddressRow>, DatabaseError>;
     fn get_scan_addresses_by_addresses(&mut self, addresses: Vec<String>) -> Result<Vec<ScanAddressRow>, DatabaseError>;
     fn add_scan_addresses(&mut self, values: Vec<ScanAddress>) -> Result<usize, DatabaseError>;
+    fn upsert_scan_addresses(&mut self, values: Vec<ScanAddress>) -> Result<usize, DatabaseError>;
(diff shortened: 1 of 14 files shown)


core/crates/storage/src/repositories/scan_addresses_repository.rs
```diff
@@ -33,6 +34,11 @@ impl ScanAddressesRepository for DatabaseClient {
         let new_addresses = values.into_iter().map(NewScanAddressRow::from_primitive).collect();
         Ok(ScanAddressesStore::add_scan_addresses(self, new_addresses)?)
     }
+
+    fn upsert_scan_addresses(&mut self, values: Vec<ScanAddress>) -> Result<usize, DatabaseError> {
```
gemcoder21: can we avoid doin upsert? just do add - and do nothing if exist already

gemcoder21: otherwise it would keep changing updated_at on each run

PR #1172, opened 2026-09-07
Seed wallet prices in every currency

DRadmir:
A new wallet only got prices when the currency was USD. For any other currency the first price fetch was thrown away because the app had no exchange rate yet, so the wallet screen stayed without prices until the live price stream connected, and never if it did not.

The price endpoint now returns the exchange rates together with USD prices, and the app stores the rate before converting, so prices show up right after the wallet is created in any currency. The API change is backward compatible; deploy it before the app release.


gemcoder21:
Remove calling this endpoint

PR #1184, opened 2026-09-08
Cap the auto formatter at four significant digits

DRadmir:
Small amounts on 18-decimal assets, such as a BNB network fee, were shown with up to 18 digits because the auto formatter fell back to full precision below 0.0001. That made fees and dust amounts unreadable on both apps.

The auto style now keeps four significant digits for those values on iOS and Android. The screens that were picking different styles on each platform are aligned as well: transaction details, delegation details, the stake minimum amount, and the amount and swap errors all use the auto style now.

Screenshot_1788854797
Screenshot_1788854808


gemcoder21 (approved):

gemcoder21 (approved):
consistent on both iOS and Android?

DRadmir:
@gemcoder21 yes, now fully consistent

PR #1185, opened 2026-09-08
Build the asset market rows in Core

DRadmir:
Core now decides which rows the asset chart's market section shows: market cap with its rank badge, fully diluted valuation, volume, contract, supplies and all-time high and low. Both apps format and localize the rows Core returns instead of each keeping its own copy of the rules.

`AssetMarket` and `ChartValuePercentage` cross the FFI as typed records instead of JSON.

On Android the market section is now built in a factory off the main thread and the scene only renders it.


gemcoder21:
Reviewed and rebased onto main (the perpetual, fiat and confirm-serde changes landed since this branched; the conflicts were all in generated files and I resolved them by regenerating).

Verified locally on the rebased commit: `core just test` + `just lint`, `just android test`, `ios just test` all green.

One adjustment pushed: `AssetChartViewModel` used `flowOn(Dispatchers.Default)` for the `marketRows` call. That call asks the explorer for the contract link, which reads the selected explorer name out of the preferences store — so it is disk I/O, not the CPU work `Default` is for, and ARCHITECTURE.md §5 wants `Dispatchers.IO` there (the repo has 55 `flowOn(Dispatchers.IO)` against 7 `Default`).

Everything else checks out. Coverage moved rather than vanished — the contract-row presence rules the deleted `AssetChartSceneTest` covered are now Rust tests, and the formatting is covered by the new `AssetMarketUIModelFactoryTest`. `AssetMarket`'s six `#[typeshare(skip)]` raw all-time fields map to nil app-to-core, which is safe: nothing in gemstone reads them, only the backend `prices`/`storage` crates do.

gemcoder21:
Second pass, on the point that the chart scene should behave like the asset scene and be ready to view without jumps.

It was not: `marketUIModel` and `title` both started empty, so opening the chart for an asset the wallet already holds rendered an empty screen and then jumped as the title and all four market sections appeared at once. `AssetDetailsViewModel` and `AssetSceneViewModel` already solve this by seeding the observation with what is in memory, so I made the chart do the same:

- Android: `AssetChartViewModel` reads the asset out of the `GetWalletAssets` StateFlow and builds the same model for the `stateIn` initial value, through one shared `marketUIModel(...)` function so the seed and the flow cannot drift — the shape `AssetDetailsViewModel` uses for its `uiModel`. Core returns the contract section with no market data (`test_market_rows_without_market_keep_contract` covers that), so the first frame is populated.
- iOS: `ChartSceneViewModel` seeds `priceQuery` with `PriceData.with(asset:)` from the asset the scene was opened for, the way `AssetSceneViewModel` seeds `assetQuery` with `ChainAssetData`.

The bootstrap tests now assert the initial value rather than `first { it != null }`, which passed either way. I mutation-checked the new one — reverting the seed to `null` fails `a stored asset gives the scene its rows and title before any flow emits`.

Verified again locally: Core tests + lint, Android suite, iOS suite green.

gemcoder21 (approved):
Reviewed, rebased and verified locally on this exact HEAD: Core tests + lint, Android suite, iOS suite all green. Three adjustments pushed to the branch — the IO dispatcher for the store-reading market rows call, first-frame seeding so the chart scene matches the asset scene, and bootstrap tests that assert the initial value instead of passing either way.

PR #1187, opened 2026-09-08
Decide what a typed amount means in Core

DRadmir:
Each app interpreted the amount field on its own: whether the text is in asset or fiat units, how it converts through the price, what Max means, when to show the reserved-fee note and what to show under the field. The two copies had already drifted: on Android, tapping Max while in fiat mode kept the fiat units and read the crypto maximum as a fiat amount, so a smaller amount went to the confirm screen with no error.

Core now answers all of it in one call, and Max always enters the amount in asset units on both platforms. The apps keep only the text field, the unit toggle, locale handling and formatting; the converters, validators and parse helpers that duplicated the rule are removed on both sides.


gemcoder21:
Reviewed and rebased onto main (the perpetual, fiat and confirm-serde changes landed since this branched; it rebased cleanly).

Verified locally on the rebased commit: `core just test` + `just lint`, `just android test`, `ios just test` all green, and `just generate-models` produces no diff, so the checked-in generated output is current.

The bug fix is the good part and it is covered: `onMaxAmount in fiat mode enters the max in asset units` switches to fiat, taps Max, and asserts the input type flips back to asset, the text is the asset maximum, and confirm receives the full atomic value with `isMax = true`. `max_entry()` returning the input type rather than just the value is the right shape — the app sets its toggle from Core's answer instead of knowing the rule.

One adjustment pushed: `AmountField.kt` had sorted imports on main and the in-place swap of `AmountInputType` for `uniffi.gemstone.GemAmountInputType` broke the order. Re-sorted. (The two other unsorted files this branch touches were already unsorted on main, so I left them.)

Two things I checked and left alone:
- `switchInputType` clearing the text matches both the pre-PR Android behaviour and iOS's `cleanInput()`, so it is not a regression.
- `buttonState` keeps its `value > 0` guard on top of `error == null`, which is needed: empty text gives `value: None, error: None`, so `error == null` alone would enable the button.

Worth a follow-up, not blocking: `AmountPerpetualProvider` and `AmountAutocloseSheet` still parse their input app-side with `parseInputNumberOrNull()`. The transfer amount is Core's now; the perpetual amounts are the remaining copy.

gemcoder21:
Rebased again onto main, which had moved.

Note the rebase was textually clean but semantically broken: `2074524376 Allow full perpetual reductions below the order minimum` landed on main a test calling `GemAmountType::validate`, the method this branch replaces with the free `validate` the entry path uses. Ported it — same assertions, now going through `validate(asset, value, available, minimum)`, with the two errors naming their asset as they do here.

Worth flagging separately: main's iOS test build was broken when I picked this up, from two changes whose iOS CI was cancelled (`a755e7d9f5` using `GemStakeServiceMock` without importing `GemstonePrimitivesTestKit`, and `d27ba3e399` adding `networkFeeBuyAmount` to `ConfirmInfoSheetBuilder.build` without updating its test). Fixed on main in `63bb14a0a0`, so this branch and #1185 build again.

Re-verified on the rebased HEAD: Core tests + lint, Android suite, iOS suite green, and `just generate-models` produces no diff.

gemcoder21 (approved):

gemcoder21 (approved):
Reviewed, rebased onto main after #1185, Core tests + lint green locally and generated output current. Two adjustments pushed: the amount field import ordering, and the perpetual-reduce test ported onto the entry API after main added a caller of the removed GemAmountType::validate.

PR #1190, opened 2026-09-08
Scanner improvements

### LINKED ISSUE #1176: Import QR Code image needs improvement
- Better way to handle an error - maybe show scan and show toast message and vibration? need consistent with android too?
- Try again does nothing (broken)

Image

gemdev111:
A photo without a QR code now shows a toast over the scanner instead of a dead-end error screen, and a photo with a code is handed over as soon as it decodes. Same on iOS and Android.

The camera-permission and no-camera screens share one layout on both apps: icon, text, actions as buttons at the bottom, photo import always available.

consolidates ios & android

Fixes #1176

images:

Simulator Screenshot - Gem-QR-1176 - 2026-09-09 at 14 03 40
Simulator Screenshot - Gem-QR-1176 - 2026-09-09 at 14 03 54


gemcoder21:
need to fix size of the button. I think we should go back and show camera view once an error is shown? Because if the button show up it would not be consistent 

Screenshot 2026-09-09 at 1 40 55 AM

gemcoder21 (approved):

PR #1204, opened 2026-09-10
Block solana pay if validation error

### LINKED ISSUE #1026: Solana Pay can hide merchant-controlled instructions behind an empty confirmation
Split out of #1015.

- **Impact:** Both platforms. A malicious Solana Pay merchant can supply a single-signer transaction with unclassified instructions. `prepare()` enforces one signer, the wallet fee payer, and a blank signature, but does not restrict instruction programs or effects. When `simple_transfer()` returns `None`, Core classifies the transaction as `SmartContractCall`; both apps preserve the encoded transaction while constructing an empty-recipient, zero-value confirmation.
- **Where:** `core/crates/payment/src/solana_pay/transaction.rs::prepare`, with confirmation fallbacks in `android/app/src/main/kotlin/com/gemwallet/android/PaymentNavigation.kt::linkRoutes` and `ios/Packages/PrimitivesComponents/Sources/Types/PaymentDestinationBuilder.swift::build`.

**Still reproducible in the tree at `b73176935c`** — verified while splitting #1015.

- **Decision:** Do not restrict Solana Pay to simple transfers. The spec defines a transaction request as "an interactive request for any Solana transaction" and only constrains signers, fee payer, and blockhash; commerce providers route payments through their own programs. Keep arbitrary instructions, make the confirmation honest, and fail closed only where payment semantics make that safe. Instruction-level inspection is tracked in #1115; the Sui severity mismatch in #1116.
- **Confirmed:** The confirmation is not empty today, but it is misleading in three ways. Core simulates the encoded transaction in `core/gemstone/src/services/confirm/mod.rs::simulate` (a transport failure already blocks the load), and the screen shows merchant, sender, network, simulated balance changes, and fee. The gaps:
  1. An execution error is `SimulationSeverity::Warning` (`core/crates/gem_solana/src/provider/simulation_mapper.rs::simulation_error_warning`), so the button stays enabled, and Solana broadcasts generic input with `skip_preflight: true` (`services/confirm/rules.rs::broadcast_options`). The merchant controls the transaction and can force this on purpose.
  2. A transaction with no signer balance delta renders nothing in the simulation section.
  3. Once loaded, the header falls back to the zero transfer value because `Generic` maps to `GemTransactionHeaderKind::Amount` (`core/gemstone/src/services/transfer/rules.rs::header_kind`). iOS hides the recipient row for generic input (`ConfirmTransferSceneViewModel.detailItems`); Android renders the zero `AmountListHead` after the `Prepare` state (`ConfirmScreen.kt`).

**Plan**

1. **Fail closed for Solana Pay only.** In the confirm service, when `metadata.source == Payment` and the transaction is not a decoded transfer, reject the load if simulation reports an execution error or no outgoing balance change for the signer. This is a single merchant-controlled transaction with no batch dependency, and a payment that moves nothing out of the wallet is worthless to the user, so there is no false-alarm cost. The error names the merchant domain. Drop `skip_preflight` for `Payment` source at the same time. WalletConnect keeps its non-blocking behavior.
2. **Header.** When a `Generic` input has zero value and no simulation header, use the asset image or symbol header kind instead of `Amount`, so "0 USDC" is never rendered. Core rule change in `header_kind`; both apps already render those kinds.
3. **Simulation section.** For externally supplied transactions, always render one of three neutral states: the balance changes, "No balance changes detected", or "Simulation unavailable". Plain informational rows in the existing balance-changes section, not warning banners; the warning banner stays reserved for execution errors. The WalletConnect path currently collapses a transport failure into an empty result (`core/gemstone/src/transaction_simulation.rs::simulate_send_transaction`, `unwrap_or_default`); surface that as the "unavailable" state instead of silence. Non-blocking.
4. **Tests.** Hostile fixtures for a forced execution error, an `Approve`-only transaction, and a real transfer combined with a hidden `Approve`; assert the payment rule rejects the first two and #1115 flags the third. Verify the new header and rows on both apps with a Solana Pay link and a WalletConnect Solana request.

0xh3rman:
Closes #1026

Simulator Screenshot - iPhone 17 Pro - 2026-09-10 at 11 35 52

### DIFF
--- a/core/gemstone/src/services/confirm/mod.rs
+++ b/core/gemstone/src/services/confirm/mod.rs
@@ -316,11 +316,13 @@ impl GemConfirmService {
         let Some(transaction) = input.transfer.input_type.simulation_payload() else {
             return Ok(None);
         };
-        self.simulation
+        let simulation = self
+            .simulation
             .simulate_transaction(chain, transaction, Some(input.from.address.clone()))
             .await
-            .map(Some)
-            .map_err(|error| GemConfirmError::Load { msg: error.to_string() })
+            .map_err(|error| GemConfirmError::Load { msg: error.to_string() })?;
+        input.transfer.input_type.validate_simulation(&simulation)?;
+        Ok(Some(simulation))
     }
 }
 
--- a/core/gemstone/src/services/confirm/rules.rs
+++ b/core/gemstone/src/services/confirm/rules.rs
@@ -1,6 +1,6 @@
 use primitives::{
     ApplicationMetadataSource, Asset, AssetId, Chain, ChainType, FeePriority, FeeUnitType, ScanAddressTarget, ScanTransaction, ScanTransactionPayload, SimulationResult,
-    Transaction, TransactionPreloadInput, Wallet,
+    SimulationWarningType, Transaction, TransactionPreloadInput, Wallet,
 };
 
 use super::error::GemConfirmError;
@@ -73,6 +73,7 @@ pub(super) trait ConfirmInput {
     fn approval_value(&self) -> Option<(AssetId, GemApprovalValue)>;
     fn validate_approvals(&self, transactions: &[GemSignedTransaction]) -> Result<(), GemConfirmError>;
     fn simulation_payload(&self) -> Option<String>;
+    fn validate_simulation(&self, simulation: &SimulationResult) -> Result<(), GemConfirmError>;
     fn broadcast_options(&self) -> GemBroadcastOptions;
 }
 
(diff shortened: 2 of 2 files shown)


gemcoder21 (approved):

core/crates/gem_solana/src/program/error.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use super::log::error_origin;
+
+pub fn error_message(error: &Value, logs: &[String]) -> Option<&'static str> {
+    let code = error.get_value("InstructionError").ok()?.at(1).ok()?.get_i64("Custom").ok()?;
+    let code = u32::try_from(code).ok()?;
+    known_program_error(error_origin(logs, code)?, code)
+}
+
+fn known_program_error(program: &str, code: u32) -> Option<&'static str> {
```
gemcoder21: that's overkill, why we need it?

gemcoder21 (commented):
need something simpler, seems overkill

Issue #1202, opened 2026-09-09
Core: architecture and build-speed plan (one artifact set, typed errors, unified chain-crate and field conventions)


gemcoder21:
## Summary

An audit of `core/` (81 workspace packages, ~250k Rust lines, 2,627 files) to find what makes it slow to build, test, and change, and what keeps the crates from sharing one shape. The headline findings:

1. **Raw compile speed is not the problem; artifact churn is.** A cold `cargo build --workspace` is 68 s wall on a 16-core Apple Silicon machine. But `cargo build`, `cargo clippy`, `cargo test --workspace`, `just test <crate>`, and `cargo build -p gemstone` each resolve a *different* feature set for the same shared crates, so the workspace is compiled up to four times into an 18 GB target dir, and switching between commands recompiles `storage`, `gemstone`, `swapper`, `settings_chain`, and `primitives` from scratch.
2. **`primitives` is a 22k-line root that 63 crates wait on, and ~26% of it is only used by one consumer** (block explorers, payment decoding, backend config keys, node checks, push models). Every edit there rebuilds 65 crates.
3. **The typed-error contract in the docs is not what the code does.** `chain_traits` returns `Box<dyn Error + Send + Sync>` from all 41 methods, gemstone's `GemstoneError` has four string escape hatches, and `GemServiceError` carries only `{ msg: String }`. 2,189 `Box<dyn Error>` sites workspace-wide.
4. **The 15 chain crates follow three client shapes, two `ChainTraits` placements, four testkit layouts, and inconsistent feature names**, with the same helper copy-pasted 8 to 15 times and at least one semantic divergence (`Failed` vs `Reverted` for the same condition).
5. **The iOS release Core build takes 22.6 min per CI run and runs on every PR**; Android release uses a different (default) profile because the release profile lives in an iOS shell script instead of `Cargo.toml`.

Everything below is grouped as the request asked: **performance**, **architecture**, **unified types and conventions**, and **overall practices**, followed by a sequencing proposal. Numbers are measured or counted; see "How this was measured" at the end.

---

## Baseline measurements

Machine: Apple Silicon, 16 cores, 64 GB, rustc 1.97.1, kache disabled (`RUSTC_WRAPPER=""`), fresh target dir, `--locked`.

| Scenario | Wall | Units / crates compiled | Notes |
|---|---|---|---|
| `cargo build --workspace` cold | 68 s | 893 units, 588 CPU-s | workspace crates are only 29% of CPU time |
[...]

PR #1206, opened 2026-09-10
Improve gemstone rust build process

0xh3rman:
Phase 0 of #1202: A3, A5, A7, and the easy part of A8.

## What improved

**A3. One release profile for both stores.** `core/Cargo.toml` defines `[profile.mobile]` (inherits `release`; fat LTO, one codegen unit, stripped symbols). `ios/scripts/generate-stone.sh` and `android/gemstone/build.gradle.kts` both select it with `--profile mobile`, so the two apps ship the same optimisation of the same crate instead of iOS getting LTO from shell env vars and Android getting Cargo defaults. `panic` stays `unwind` so UniFFI keeps turning Rust panics into foreign exceptions. Backend containers keep plain `--release`. Release outputs now live under `target/<triple>/mobile/`; the two Xcode Release link paths and `GEMSTONE_IOS_PROFILE` follow. Rationale recorded in `docs/ARCHITECTURE.md` § 14.

**A5. `gemstone` is an rlib by default.** `crate-type = ["rlib"]`; the iOS recipes pass `cargo rustc --crate-type staticlib`, the Android task and the two bindgen recipes pass `--crate-type cdylib`, and the Android host-test recipe (`just android build-gemstone-host`) does the same so JNA still finds `target/debug/libgemstone.dylib`. Host `cargo build` and `cargo test` no longer link a static archive and a dylib of the whole dependency tree on every run. The unused `[build-dependencies] uniffi` entry is gone (there is no `build.rs`). cargo-ndk 4.1.2 reads crate types from cargo's JSON artifact messages, and cargo reports the override as `cdylib`, so the `-o jniLibs` copy still works.

**A7. Tests run through cargo-nextest.** `just test`, `just test <crate>`, and `just test-integration` use `cargo nextest run` with filtersets instead of `cargo test` filters. Test binaries run in parallel and the workspace suite now executes in about 11 s where the sequential run was 88 s. `just install-nextest` fetches the pinned 0.9.143 prebuilt (macOS or Linux) and is part of `just install`; the Core CI unit-test job installs it before `just test`. Nextest fails when a selection matches no tests, which turns the "compiled nothing, passed" case into a failure.

**A8, partial.** `ed25519-dalek` in `signer` and `gem_aptos` and `bcs` in `gem_sui` now use the workspace entries. The `gas-bench` tool is removed outright (the issue's `prettytable-rs` finding was wrong, the tool imported it; the tool itself had no consumers), which prunes it and 12 exclusive dependencies from the lock. `uniffi_build` leaves the lock with the build-dependency.

## Verification

- `just test`: 3121 passed, 1 skipped (pre-existing ignore), 82 binaries
- `just test gem_bitcoin`: 58 passed, 9 integration tests filtered; `cargo nextest list -p fiat --features fiat_integration_tests -E 'test(fiat_integration_tests)'` lists 16
- `just lint`; `cargo clippy -p signer -p gem_aptos -p gem_sui --all-features --all-targets -- -D warnings`
- `cargo build -p gemstone --locked` produces the rlib only; `cargo metadata --locked --offline` resolves
- `cargo ndk -t arm64-v8a -o … rustc --lib --crate-type cdylib` and `./gradlew :gemstone:buildCargoNdk` copy `libgemstone.so`
- `cargo rustc --target aarch64-apple-ios-sim --profile mobile --crate-type staticlib`: 5m32s, no warnings, artifact at the new path
- `just ios generate-stone` (debug) regenerates bindings and builds both static libraries

Not run: a full Xcode Release archive and an Android release bundle, so the Release link path and AGP handling of the pre-stripped `.so` are verified at the artifact level only. Integration recipes were listed, not executed.

## Left for later

Version dedup, `cargo deny`, the remaining inline-versioned dependency lines, thin-vs-fat LTO measurement, and `--timings` in CI (A4).


gemcoder21:
@0xh3rman what does this improve?

PR #1221, opened 2026-09-10
Let a failed Monad staking call fail instead of reporting no delegations

DRadmir:
When the Monad staking call failed, the wallet read the empty result as the truth and deleted every stored Monad delegation, so the stake screen showed a staked balance with nothing to unstake or claim until a later sync succeeded.

Now the failure is reported instead of being hidden, which is what every other staking chain already does. A wallet that never staked still gets an empty list without an error.


0xh3rman (approved):
Is this caused by network disconnected? not node issue

DRadmir:
@0xh3rman not only a disconnect, a node error or rate limit hits the same path. Any of them used to read as "no delegations" and wipe the stored ones

PR #1225, opened 2026-09-11
Count the Solana unstake wait from the epoch's progress

DRadmir:
A Solana stake that is activating or deactivating becomes available when the current epoch ends. The wait was measured against the epoch number instead of how far the epoch had actually run, so it read about two days for the whole epoch, no matter how close the end was.

Checked on a watch wallet with a real deactivating stake: before it said 2 days 2 hours, after it said 19 hours 59 minutes, which is what the chain reported at that moment.

Before

01-before-list
03-before-detail

After

02-after-list
04-after-detail

### DIFF
--- a/core/crates/gem_solana/src/provider/staking_mapper.rs
+++ b/core/crates/gem_solana/src/provider/staking_mapper.rs
@@ -1,5 +1,5 @@
 use crate::models::{EpochInfo, TokenAccountInfo, VoteAccount};
-use chrono::Utc;
+use chrono::{DateTime, Duration, Utc};
 use num_bigint::BigUint;
 use primitives::{AssetId, Chain, DelegationBase, DelegationState, DelegationValidator};
 
@@ -24,7 +24,7 @@ pub fn map_staking_validators(vote_accounts: Vec<VoteAccount>, chain: Chain, net
         .collect()
 }
 
-pub fn map_staking_delegations(stake_accounts: Vec<TokenAccountInfo>, epoch: EpochInfo, asset_id: AssetId) -> Vec<DelegationBase> {
+pub fn map_staking_delegations(stake_accounts: Vec<TokenAccountInfo>, epoch: EpochInfo, asset_id: AssetId, now: DateTime<Utc>) -> Vec<DelegationBase> {
     stake_accounts
         .into_iter()
         .filter_map(|account| {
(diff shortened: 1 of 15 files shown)


core/crates/gem_solana/src/provider/staking_mapper.rs
```diff
@@ -123,8 +121,8 @@ mod tests {
                             token_amount: None,
                             stake: Some(crate::models::StakeInfo {
```
0xh3rman:
```suggestion
                            stake: Some(StakeInfo {
```
import and use

core/crates/gem_solana/src/provider/staking_mapper.rs
```diff
@@ -135,19 +133,40 @@ mod tests {
                 owner: "owner1".to_string(),
                 lamports: 1000000,
             },
-        }];
+        }
+    }
 
-        let epoch = EpochInfo {
+    fn epoch_info(slot_index: u64) -> EpochInfo {
```
0xh3rman: move to `EpochInfo` as a new initializer

core/gemstone/src/services/stake/rules.rs
```diff
@@ -449,6 +457,34 @@ mod tests {
     use super::*;
     use primitives::{AssetId, Resource};
 
+    fn solana_validator(name: &str) -> DelegationValidator {
```
0xh3rman, line 460: this is a mock helper? move to testkit or mock if available

core/gemstone/src/services/stake/rules.rs
```diff
@@ -67,6 +67,13 @@ pub fn can_claim_rewards(wallet_type: WalletType, delegation: &Delegation) -> bo
     wallet_type != WalletType::View && config.can_claim_rewards && shows_rewards(&delegation.base)
 }
 
+pub fn validator_display_name(validator: &DelegationValidator) -> String {
+    if validator.name.trim().is_empty() {
```
0xh3rman: where does this name come from? maybe no need to trim?

0xh3rman (approved):
lgtm overall, left some comments

ios/Features/Stake/Tests/ViewModels/ValidatorViewModelTests.swift
```diff
@@ -11,4 +12,11 @@ struct ValidatorViewModelTests {
 
         #expect(model.aprModel.text == "APR 2.15%")
     }
+
+    @Test func nameFallsBackToTheAddressWhenUnnamed() {
+        let validator = DelegationValidator.mock(.solana, id: "8GbwASqdpw4dVcwbWUxbHXMrjyQx2aKkoBR5H1GJF8iD", name: "")
+
```
gemcoder21, line 18: unit test is not needed, it's tested on gemstone side

PR #1250, opened 2026-09-15
Add bitcoin support for relay

0xh3rman:
Add Bitcoin as a Relay source chain using Relay deposit addresses.

1. Quote with `useDepositAddress` for Bitcoin; the deposit step returns a per-request `depositAddress` and the deposit `amount`, no PSBT and no OP_RETURN memo.
2. Map the quote to a transfer-type `SwapQuoteData` to the deposit address, signed by the existing Bitcoin planner; no new signer code.
3. Use `SwapAmountMode::Flexible` for Bitcoin: open deposit addresses re-quote for the deposited amount, so Max sweeps like THORChain and no `RESERVED_NATIVE_FEES` entry is needed.
4. Validate the deposit step at quote time and again when building quote data, dispatched per `RelayChain` so `get_quote` stays chain-agnostic.
5. Register Relay chain id 8253038 and the Bitcoin currency sentinel; map swap results and vault addresses for Bitcoin.
6. Add X Layer to the Relay chain list and update the FEATURES tables.

Verified live: BTC → Base USDC quote returns a `bc1q` deposit address with a 5,000 sat gas fee, and the request lookup by Bitcoin txid resolves a completed swap (`test_relay_bitcoin_live`).


core/crates/gem_bitcoin/src/provider/preload.rs
```diff
@@ -47,12 +47,16 @@ impl<C: Client> ChainTransactionLoad for BitcoinClient<C> {
...
-    async fn get_transaction_fee_rates(&self, _input_type: TransactionInputType) -> Result<Vec<FeeRate>, Box<dyn Error + Sync + Send>> {
+    async fn get_transaction_fee_rates(&self, input_type: TransactionInputType) -> Result<Vec<FeeRate>, Box<dyn Error + Sync + Send>> {
         match self.chain {
             BitcoinChain::Bitcoin | BitcoinChain::Litecoin | BitcoinChain::BitcoinCash | BitcoinChain::Doge => {
                 let priority = self.chain.get_blocks_fee_priority();
                 let (slow, normal, fast) = futures::try_join!(self.get_fee(priority.slow), self.get_fee(priority.normal), self.get_fee(priority.fast))?;
-                Ok(map_fee_rates(slow, normal, fast, self.chain))
+                let mut rates = map_fee_rates(slow, normal, fast, self.chain);
+                if input_type.is_contract_swap() {
+                    rates.retain(|rate| rate.priority == FeePriority::Normal);
```
gemcoder21: let's avoid muts and we should let user change it

gemcoder21 (commented):
can you also increase bitcoin timeout to 3 hours for other chains? for swaps? we had 1 hour I think?

0xh3rman:
> can you also increase bitcoin timeout to 3 hours for other chains? for swaps? we had 1 hour I think?

do you mean `swap_transaction_timeout`?

core/crates/swapper/src/fees/reserve.rs
```diff
@@ -8,6 +8,7 @@ use crate::{QuoteRequest, SwapperError};
 
 pub static RESERVED_NATIVE_FEES: LazyLock<HashMap<Chain, &'static str>> = LazyLock::new(|| {
```
gemcoder21, line 9: `RESERVED_NATIVE_FEES` should not be used and removed

PR #1257, opened 2026-09-15
Own the app lock in Core and authenticate through one door on iOS

DRadmir:
Confirming a transaction with Face ID showed the Privacy Lock screen over the app for the whole prompt, and the lock timer could ask for Face ID after a one-second background trip once the app had been open for longer than the lock period.

The app lock now lives in Core as one session used by both apps: when to lock, when to cover the screen, when to prompt, and what to do when a prompt is cancelled or interrupted. On iOS every authentication prompt, including wallet creation and import, goes through a single door, so the lock screen always knows the prompt is the app's own.


gemdev111 (approved):

gemcoder21:
@DRadmir I wonder if we should handle it differently? only show privacy lock when you go on the background and ignore inactive state? I checked Phantom, they do that

PR #1299, opened 2026-09-17
Never leave an imported wallet without its secret

DRadmir:
An import that failed after the wallet row had already been written deleted only the secret, so the wallet stayed in the list with nothing behind it: it could not sign, could not show its phrase, and re-importing the same phrase was refused as already imported — leaving no way out but deleting it first.

The import now removes the wallet row as well, and removes the secret only once the row is gone, so a failed import leaves nothing behind and can simply be retried.


gemcoder21:
Any better ways to handle it? to either do that after or in the transaction?

0xh3rman:
How to reproduce this case? only delete the keystore file from FS?

DRadmir:
@0xh3rman failure arises from the import, the wallet was created, but the address names thrown error. The old rollback only erased the secret and left the wallet

gemcoder21:
> @0xh3rman failure arises from the import, the wallet was created, but the address names thrown error. The old rollback only erased the secret and left the wallet

But why did it fail? what was the error / an issue?

